PR #9, opened 2020-06-15
Swagger docu linking

FabiKo117:
adding links to swagger in docu and vice-versa
adding and modifying of further links to javadoc and github

### DIFF
--- a/docs/filter.rst
+++ b/docs/filter.rst
@@ -151,4 +151,6 @@ Here's some useful examples for querying some OSM features:
 Further Information
 -------------------
 
-The filter parameter is powered by a separate Java module, which can be used independently as a maven library. Find the open source code here: https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/libs/ohsome-filter#readme. The corresponding Javadoc documentation is available at: https://docs.ohsome.org/java/ohsome-filter/
+The filter parameter is powered by a separate Java module, which can be used independently as a maven library.
+You can further find the `open source code <https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/libs/ohsome-filter#readme/>`_, 
+as well as the corresponding `Javadoc documentation <https://docs.ohsome.org/java/ohsome-filter/>`_.
--- a/docs/filter.rst
+++ b/docs/filter.rst
@@ -97,14 +97,14 @@ Here's some useful examples for querying some OSM features:
     +------------------+--------------------------------------------------------+------------------------------+
     | **OSM Feature**  | **filter**                                             | **comment**                  |
     +==================+========================================================+==============================+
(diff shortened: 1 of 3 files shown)


docs/index.rst
```diff
@@ -29,3 +29,10 @@ Welcome to the documentation of the ohsome API!
    group-by
    time
    filter
+   
+Additional Links:
+-----------------
+
+* `API Code development <https://github.com/GIScience/ohsome-api>`_
+* `Swagger API playground <https://api.ohsome.org/v1/swagger-ui.html>`_
```
joker234, line 37: Would it be better to have a `/stable` link or is it sufficient to have `/v1` since we have to change a lot of things here for a new (future) version?

docs/filter.rst
```diff
@@ -151,4 +151,6 @@ Here's some useful examples for querying some OSM features:
 Further Information
 -------------------
 
-The filter parameter is powered by a separate Java module, which can be used independently as a maven library. Find the open source code here: https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/libs/ohsome-filter#readme. The corresponding Javadoc documentation is available at: https://docs.ohsome.org/java/ohsome-filter/
+The filter parameter is powered by a separate Java module, which can be used independently as a maven library.
+You can further find the `open source code <https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/libs/ohsome-filter#readme/>`_,
```
joker234, line 155:
```suggestion
You can further find the `open source code <https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/libs/ohsome-filter#readme>`_, 
```

docs/index.rst
FabiKo117: As far as I can tell, there is no /stable link yet :)

And, if we have them, I guess we could link both: the /stable instance, as well as the /development

joker234: > As far as I can tell, there is no /stable link yet :)
No there isn't. But we can create one if that helps ;)

> And, if we have them, I guess we could link both: the /stable instance, as well as the /development
If you want to publish that (unstable) development instance, yes :thinking:

FabiKo117: Alright :) I agree, that pointing to /stable instead of a specific version would be smarter as we won't have to adapt it everytime for a new version.

So should the development instance just be for internal usage then? Thought we could also make it externally accessible, e.g. if someone needs a "quick & dirty" feature for some measurement or such. Then just linking it there (without providing a specific docu) could be useful.

joker234:
```suggestion
* `Swagger API playground <https://api.ohsome.org/stable/swagger-ui.html>`_
```

FabiKo117: Alright :)
No other change then needed I think.. will commit that suggestion and then merge as soon as the link is working.

PR #12, opened 2020-07-06
check repeated parameters

bonaparten:


bonaparten:
I suppose, we only want to return a message 40x and let the user modify/remove the repeated parameter. Therefore, I made a simple implementation that splits the query string and checks if parameters are repeated. Let me know if this is a good solution for the issue.

PR #14, opened 2020-07-09
updating swagger parameter descriptions

FabiKo117:
marking the types, keys, values params as deprecated
moving default values from deprecated params to filter
adapting some descriptions and default values


FabiKo117:
I just saw that there are also quite some not-related formatting updates. Guess they are related to the change of the package naming.

PR #15, opened 2020-07-13
Update readme docs

FabiKo117:
updating readme to include more usability infos (e.g. links to docs, instances, etc.)
adding a CONTRIBUTING.txt file that includes info how to contribute to the ohsome API

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # ohsome API
 
-This REST-based API aims to leverage the tools of the [OSHDB-API](https://github.com/GIScience/oshdb) through allowing to access some of its functionalities via HTTP requests. The current stable version is [v1.0.0](https://github.com/GIScience/ohsome-api/releases/tag/1.0.0). The code is developed and maintained by [HeiGIT](https://heigit.org/).
+This REST-based API aims to leverage the tools of the [OSHDB API](https://github.com/GIScience/oshdb) through allowing to access some of its functionalities via HTTP requests. The current stable version is [v1.0.0](https://github.com/GIScience/ohsome-api/releases/tag/1.0.0). The code is developed and maintained by [HeiGIT](https://heigit.org/).
 
 [![](http://jenkins.ohsome.org/buildStatus/icon?job=ohsome-api/master)](http://jenkins.ohsome.org/blue/organizations/jenkins/ohsome-api/activity)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,17 +1,26 @@
 # ohsome API
 
-This REST-based API aims to leverage the tools of the [OSHDB-API](https://github.com/GIScience/oshdb) through allowing to access some of its functionalities via HTTP requests.
-The current stable version is [v1.0.0](https://github.com/GIScience/ohsome-api/releases/tag/1.0.0).
+This REST-based API aims to leverage the tools of the [OSHDB API](https://github.com/GIScience/oshdb) through allowing to access some of its functionalities via HTTP requests. The current stable version is [v1.0.0](https://github.com/GIScience/ohsome-api/releases/tag/1.0.0). The code is developed and maintained by [HeiGIT](https://heigit.org/).
```
tyrasd, line 3: IMHO this very first introduction sentence could be written even more welcoming and maybe a bit less technical. There would be plenty of opportunity below to explain that it is implemented using the oshdb-api for example. Also the code ownership is not that important to be in the very first paragraph. Instead, maybe we could use this prime "screen real estate" to be a very short introduction of what the ohsome API is and what it does? For example:

> The ohsome API is a generic web API for in-depth analysis of OSM data including it's history. It allows to get aggregated statistics about the evolution of OpenStreetMap data itself and about the contributors behind the data. Furthermore, data extraction methods are provided to access the historic development of individual OSM features.

(just a first very quick first draft)

README.md
```diff
@@ -180,7 +147,7 @@ DOCS_DEPLOYMENT=release make clean html
 
 ## Built With
 
-* [Eclipse](http://www.eclipse.org/downloads/packages/eclipse-ide-java-ee-developers/oxygen1a) - IDE
+* [Eclipse](https://www.eclipse.org) - IDE
```
tyrasd, line 150: do we really need to mention the IDE? I mean, it's possible to use other IDEs just fine, and not even all of us use eclipse regularly :wink:

FabiKo117: true, in the meantime we have more collaborators that also use different IDEs :) 
I'll remove that part about the IDE.

README.md
FabiKo117: > Also the code ownership is not that important to be in the very first paragraph.

I took this idea directly from the readme of the OSHDB here on GitHub. But yeah we can add some more info and divide it into 2 paragraphs.
How about something like this?

> The ohsome API is a generic web API for in-depth analysis of OpenStreetMap (OSM) data with a focus on it's history. It allows to get aggregated statistics about the evolution of OSM data itself and about the contributors behind the data. Furthermore, data extraction methods are provided to access the historic development of individual OSM features.

> The functionalities of the ohsome API can be accessed via HTTP requests. As a basis underneath serves the OSHDB API. The current stable version is v1.0.0. Developed and maintained by HeiGIT.

tyrasd: true, it was even me who added it there: https://github.com/GIScience/oshdb/commit/8447a781356d61c3b695b31f65eb4cc6cfbfc80b :sweat_smile: maybe we should adapt the OSHDB readme file at some point as well…

I like your two-paragraph solution. :+1:

PR #18, opened 2020-07-22
WIP: adding functionality to only request changes in full-history extraction

FabiKo117:
adding 'modifications' as a value to the properties parameter
gives the full-history data without the unmodified features for the requested timespan

### DIFF
--- a/src/main/java/org/heigit/ohsome/ohsomeapi/executor/ElementsRequestExecutor.java
+++ b/src/main/java/org/heigit/ohsome/ohsomeapi/executor/ElementsRequestExecutor.java
@@ -145,7 +145,7 @@ public class ElementsRequestExecutor {
     DataResponse osmData = new DataResponse(new Attribution(URL, TEXT), Application.API_VERSION,
         metadata, "FeatureCollection", Collections.emptyList());
     try (Stream<Feature> streamResult = preResult.stream()) {
-      exeUtils.streamElementsResponse(servletResponse, osmData, false, streamResult, null);
+      exeUtils.streamElementsResponse(servletResponse, osmData, streamResult);
     }
   }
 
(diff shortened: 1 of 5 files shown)


src/main/java/org/heigit/ohsome/ohsomeapi/inputprocessing/InputProcessor.java
```diff
@@ -867,4 +870,8 @@ public boolean includeOSMMetadata() {
   public boolean isUnclipped() {
     return unclipped;
   }
+  
+  public boolean modifications() {
```
rtroilo, line 874: suggestion to keep a more consistent naming schema the getter method of the modifications boolean should be named isModifications()!

src/main/java/org/heigit/ohsome/ohsomeapi/executor/ElementsRequestExecutor.java
```diff
@@ -299,52 +300,59 @@ public static void extractFullHistory(ElementsGeometry elemGeom,
...
+          Stream<Feature> snapshotStream = snapshotPreResult.stream()) {
+        exeUtils.streamElementsResponse(servletResponse, osmData, true, snapshotStream,
+            contributionStream);
+      }
+    } else {
+      try (Stream<Feature> contributionStream = contributionPreResult.stream()) {
+        exeUtils.streamElementsResponse(servletResponse, osmData, true, null,
+            contributionStream);
+      }
     }
```
rtroilo, line 356: code duplication
Stream implements Autoclosable but it is generally not necessary to close streams at all. You only need to close streams that use IO resources, which is this case they do not.

You can avoid this duplication here by moving the declaration of contributionStream and snapshotStream before the if block

```
Stream<Feature> contributionStream = contributionPreResult.stream();
Stream<Feature> snapshotStream = null;
if (!onlyModifications) {
...
  snapshotStream = snapshotPreResult.stream();
}
 exeUtils.streamElementsResponse(servletResponse, osmData, true, 
   snapshotStream,
   contributionStream);
```

src/main/java/org/heigit/ohsome/ohsomeapi/inputprocessing/InputProcessor.java
FabiKo117: Yeah I was thinking about that one how to name it properly.. it's true that a boolean getter should somehow have "is" or "has" in it's name. I decided to not go for "hasModifications()" as the response always has the modifications in it, no matter if this property is given or not. "isModifications()" does also not fit that well in my opinion..
What about something like "hasModificationsOnly()" and then also naming the boolean "modificationsOnly" but keep the name of the property inside the request "modifications" for simplicity purpose?

rtroilo: Yes this sounds better, hasModificationsOnly() or maybe  better than this, as  this is a option of the query and you could ask for it like, "is this query for modifications only" , isModificationsOnly().

tyrasd:
I don't particularly like that this is mixed with the `properties` parameter. I would not expect this to be the parameter to set, when reading only the documentation at https://docs.ohsome.org/ohsome-api/v1/endpoints.html#extraction-endpoints:

> properties – specifies what properties should be included for each feature representing an OSM element

Wouldn't it be better to have this as a new resource and or specific dedicated request property?

src/main/java/org/heigit/ohsome/ohsomeapi/executor/ElementsRequestExecutor.java
tyrasd: Why do you think this is not necessary to (auto)close the stream here? At least in some backends, the closing of the result stream would actually close some otherwise left open IO resources, e.g. here: https://github.com/GIScience/oshdb/blob/29c29056615569a45195deac60355b238ad3df7e/oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/backend/MapReducerIgniteScanQuery.java#L545-L547

rtroilo: That's a good point, and I would guess, that the majority of users would never close the stream like me :-).
So we should write a warning in the doc that this is a resource stream and should be close.

For this request I suggest,as it was to pull the `snapshotPerResult` for the if block again and check it in the try-with-resources later
```
try (Stream<Feature> snapshotStream = (snapshotPreResult != null)?snapshotPreResult.stream():null;
     Stream<Feature> contributionStream = contributionPreResult.stream()) {
  exeUtils.streamElementsResponse(servletResponse, osmData, true, snapshotStream, contributionStream);
}
```

FabiKo117:
> I don't particularly like that this is mixed with the `properties` parameter. I would not expect this to be the parameter to set, when reading only the documentation at https://docs.ohsome.org/ohsome-api/v1/endpoints.html#extraction-endpoints:

The documentation was also not up-to-date as the 'unclipped' feature was missing there, which I've added now as well in here for swagger and for the docs.

> Wouldn't it be better to have this as a new resource and or specific dedicated request property?

Don't think it would fit that well as a new resource as it's something full-history specific. It could be extracted as a standalone parameter though. In fact it differs from the other 3 properties (tags, metadata, unclipped) in a way that it filters out features, instead of just adding/removing/modifying the existing features.

And another thing: I've re-labelled the pr to WIP, because we also want to add the functionality to only get the latest changes of the objects, which would be required for the analysis of @Zia-. This would mean that we would need 2 new parameters for the /fullHistory resource: "modificationsOnly" and "lastModificationOnly".

src/main/java/org/heigit/ohsome/ohsomeapi/executor/ElementsRequestExecutor.java
rtroilo: Related to the MapReducerIgniteScanQuery I found a nice article, that the flatMap operation will close the inner stream in any case.
https://rmannibucau.metawerx.net/post/did-you-think-about-closing-your-java-stream section "java.util.stream, close and flatMap"
And luckily this is exactly how we handle the returning Stream of `mapStreamOnIgniteCache`

https://github.com/GIScience/oshdb/blob/29c29056615569a45195deac60355b238ad3df7e/oshdb-api/src/main/java/org/heigit/bigspatialdata/oshdb/api/mapreducer/backend/MapReducerIgniteScanQuery.java#L198

tyrasd: thanks, I didn't know about this! good to know. I have to say though that I find it quite strange that they did implement this "automatic closing of a stream when it is consumed" for `flatMap`'ed streams, but not regular streams. :shrug:

tyrasd:
> The documentation was also not up-to-date as the 'unclipped' feature was missing there, which I've added now as well in here for swagger and for the docs.

If I remember correctly, we talked about this a a few months ago and decided that we do not document the `properties=unclipped` feature _on purpose_. It was because it does not fit logically into the parameter, and we also wanted to create a new dedicated parameter for it. I can't find a ticket for it, however. I will create one now.

tyrasd:
> Don't think it would fit that well as a new resource as it's something full-history specific.

I can't follow… Just because two resources are in some way similar doesn't mean they can't exist side by side? I mean, we have for example `/elements/count` and `/elements/length`, which are both not-full-history specific, yet they do slightly different things and have a place next to each other. We could have implemented the `/count` and `/length` part of the resource easily also as a `?aggregateMethod=count` or `?aggregateMethod=length` query search parameter, but we chose not to.

Similarly, we could choose to not squeeze too much functionality into a single generic full-history resource, but have multiple resources side by side which do slightly different things. IMHO, this could work nicely:

* `/elementsFullHistory/geometry/` for the real full history data extraction
* `/elementsFullHistory/contributions/geometry/` for the modifications only 
* `/elementsFullHistory/contributions/latest/geometry/` for the "latest" changes only

we could later maybe add some further additional stuff like:

* `/elementsFullHistory/contributions/count` for count-based statistics for modifications
* `/elementsFullHistory/geometry/difference` for a difference between two points in time
* …

to me, these longer, but more verbose resource paths make it much clearer what the different modes of operation do, compared to hide the complexity in small flags.

Another problem with flags is that they might complicate and hinder further expansion, e.g. when you add one more different mode of operation, you always have to take into consideration how it will interact with all other possible flags that are already implemented. If you limit a certain feature to it's own resource, you are more free to expand and grow the whole system without having too much about interaction of different code branches.

For example, I would think that for the `lastModificationOnly` "mode" is a good example for this: I could imaging, that it would potentially be quite hard to find a way to implement the geojson output for it that works in an intuitive way: How would you encode deleted elements? A deletion is clearly a modification, but one without a geometry that can be used for the corresponding GeoJSON feature. So, if the last modification of a feature was a deletion, you either need to return features with no (or `null`) geometry (which many software tools don't support), or fall back to return the modification before the last one, or maybe it's better to return the result not in GeoJSON format, but some structure where you can say explicitly say what type of modification a chage is (somewhat similar to https://wiki.openstreetmap.org/wiki/OsmChange perhaps, but using JSON instead of XML).

rtroilo:
I like the idea to have this resources side by side, too! It looks more consistend.

For deletions I would return the very last geometry at this point. So the user can, for example, highlight deleted features or can calculate the added/deleted sum of all modifications, ... .

tyrasd:
> For deletions I would return the very last geometry at this point. So the user can, for example, highlight deleted features or can calculate the added/deleted sum of all modifications, ... .

This would be an option, but if a user just wants to update an extract to the most current state of OSM/ohsome, they only need the information that the feature with this type/id was deleted, and anything else would be superfluous. It depends I guess. ;)

FabiKo117:
closing this PR as the requested features will be implemented in a different branch and merged through PR #34.

Issue #17, opened 2020-07-17
Docs: add possibility to copy curl requests


joker234:
On some example pages one could copy an example request (especially those that are directly usable in shell) to clipboard by clicking on a button. That might be useful for the docs as well.
e.g. for example here: https://docs.ohsome.org/ohsome-api/stable/endpoints.html#post--elements-(aggregation)

bonaparten:
solved with #21

PR #21, opened 2020-07-28
Copy url to clipboard

bonaparten:
It solves #17.
In case of example boxes with JSON, it copies the JSON content.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -48,6 +48,7 @@ extensions = [
   'sphinxcontrib.openapi',
   'sphinxcontrib.httpdomain',
   'sphinx_tabs.tabs',
+  'sphinx_copybutton',
 ]
 
 # Add any paths that contain templates here, relative to this directory.
--- a/docs/endpoints.rst
+++ b/docs/endpoints.rst
@@ -33,25 +33,25 @@ Aggregation Endpoints
 .. _time: time.html#time
 .. _filter: filter.html#filter
 
+
 **Example request**:
+
+How big is the area of farmland in the region Rhein-Neckar?
 
  .. tabs::
 
    .. code-tab:: bash curl (GET)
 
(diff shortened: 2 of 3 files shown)


joker234:
Could you please not merge the master into this branch but rebase it on master instead? It's almost impossible to review your PR using the GitHub tooling, because all changes from master are displayed too ;)

docs/conf.py
```diff
@@ -180,3 +181,7 @@
 
 # A list of files that should not be packed into the epub file.
 epub_exclude_files = ['search.html']
+
+# Sphinx-button skips lines which start with '#'
```
joker234, line 185: Only works for some windows. But nothing that needs to be changed.

docs/conf.py
```diff
@@ -48,6 +48,7 @@
   'sphinxcontrib.openapi',
   'sphinxcontrib.httpdomain',
   'sphinx_tabs.tabs',
+  'sphinx_copybutton',
```
joker234, line 51: This module is missing in the requirements. The docs build fails because of that.

docs/endpoints.rst
```diff
@@ -33,7 +33,9 @@ Aggregation Endpoints
 .. _time: time.html#time
 .. _filter: filter.html#filter
 
-**Example request**:
+**How big is the area of farmland in the region Rhein-Neckar?**
```
joker234, line 36: It's one possibility to add a header. I, personally, would be in favour of a description below “Example Request”. This current test leads to unpleasant looking indentions ;)
Clipboard - July 30, 2020 3_11 PM

bonaparten: The indention was on purpose. I wanted to show, how it could look and how it looked. "Example response" should also be moved lightly to the right side.

PR #29, opened 2020-08-19
addition of parameter ‘unclipped’

### LINKED ISSUE #20: finalize and document "unclipped geometry mode"
In v1.0, there exists an (undocumented) feature flag in the data extraction endpoints, which disables the clipping to the query area-of-interest of the geojson feature geometries.

[The currently undocumented implementation of having it mixed up into the `properties` parameter is not great, because this parameter should just alter the `properties` of the returned GeoJSON, nothing more but also nothing less than that. `unclipped` doesn't fit into this parameter, because it alters the `geometry` of the resulting GeoJSON.]

In order to complete this feature, we need to:
 
* [ ] choose and create a dedicated parameter for it
* [ ] add it to the documentation

see also https://github.com/GIScience/ohsome-api/pull/18#issuecomment-663558742

FabiKo117:
implementing processing and adding it to the logic
adding it to the docs (swagger and sphinx)
refactoring boolean parameter usage
adding test
closes #20

### DIFF
--- a/docs/endpoints.rst
+++ b/docs/endpoints.rst
@@ -1030,7 +1030,9 @@ Extraction Endpoints
 
    :query time: required; format same as described in time_
    :query properties: specifies what properties should be included for each feature representing an OSM element: ‘tags’ and/or 'metadata’; multiple values can be delimited by commas; default: empty
+   :query unclipped: boolean operator to apply a clipping to the geometries of the returned featues (‘true’), or not (‘false’); default: ‘false’
    :query <other>: see above_ (except **format**)
+   
 
 .. note:: The extraction endpoints always return a .geojson file.
 
@@ -1091,9 +1093,11 @@ Get all the bike rental stations in Heidelberg.
    This endpoint supports the same ``geometryType`` options as the ``/elements`` endpoint.
 
    :query time: required; must consist of two ISO-8601 conform timestrings defining a time interval; no default value
-   :query properties: specifies what properties should be included for each feature representing an OSM element: ‘tags’ and/or 'metadata’; multiple values can be delimited by commas; default: empty
+   :query properties: same as for generic-extraction_
+   :query unclipped: same as for generic-extraction_
    :query <other>: see above_ (except **format**)
 
+.. _generic-extraction: endpoints.html#post--elements-(geometryType)
 .. _above: endpoints.html#post--elements-(aggregation)
 .. _time: time.html#time
 
--- a/docs/endpoints.rst
+++ b/docs/endpoints.rst
@@ -1030,7 +1030,7 @@ Extraction Endpoints
 
    :query time: required; format same as described in time_
    :query properties: specifies what properties should be included for each feature representing an OSM element: ‘tags’ and/or 'metadata’; multiple values can be delimited by commas; default: empty
-   :query unclipped: boolean operator to apply a clipping to the geometries of the returned featues (‘true’), or not (‘false’); default: ‘false’
+   :query clipGeometry: boolean operator to specify whether the returned geometries of the features should be clipped to the query's spatial boundary (‘true’), or not (‘false’); default: ‘true’
    :query <other>: see above_ (except **format**)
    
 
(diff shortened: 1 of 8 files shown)


docs/endpoints.rst
```diff
@@ -1030,7 +1030,9 @@ Extraction Endpoints
 
    :query time: required; format same as described in time_
    :query properties: specifies what properties should be included for each feature representing an OSM element: ‘tags’ and/or 'metadata’; multiple values can be delimited by commas; default: empty
+   :query unclipped: boolean operator to apply a clipping to the geometries of the returned featues (‘true’), or not (‘false’); default: ‘false’
```
tyrasd, line 1033: I'm wondering if this is clear enough… maybe we can tell what the geometries will be clipped against?

tyrasd (commented):
looks good. small suggestion: maybe we should put a comment in the code that the old (undocumented) variant with `properties=unclipped` is deprecated, so we don't forget to remove it in the next release(s). perhaps it's possible to use `@Deprecated` for that?

docs/endpoints.rst
FabiKo117: You mean something like ".. apply a clipping on the returned features using the defined spatial boundary..."?

FabiKo117:
> looks good. small suggestion: maybe we should put a comment in the code that the old (undocumented) variant with `properties=unclipped` is deprecated, so we don't forget to remove it in the next release(s). perhaps it's possible to use `@Deprecated` for that?

I've tried this already, but can't find a proper position where to add the @Deprecated annotation. Can you find one maybe?

docs/endpoints.rst
tyrasd: i would have maybe written something like: _“boolean operator to specify whether the returned features' geometries should be clipped to the query's spatial boundary (‘false’), or not (‘true’); default: ‘false’”_

tyrasd: PS: the values `true`/`false` are reversed in the description above: `true` would be for unclipped features, while `false` returns the clipped ones.

tyrasd:
> > looks good. small suggestion: maybe we should put a comment in the code that the old (undocumented) variant with `properties=unclipped` is deprecated, so we don't forget to remove it in the next release(s). perhaps it's possible to use `@Deprecated` for that?
> 
> I've tried this already, but can't find a proper position where to add the @deprecated annotation. Can you find one maybe?
> Cause it's only relevant in line [464](https://github.com/GIScience/ohsome-api/blob/3628d042df46634d31e8f152354689ef1f1d5b08/src/main/java/org/heigit/ohsome/ohsomeapi/inputprocessing/InputProcessor.java#L464) in InputProcessor.java.

I believe it should be possible if one rewrites the whole `if - else if - else` statement slightly, like this:

```java
    for (String property : properties) {
      @Deprecated boolean oldUnclippedParameter = "unclipped".equalsIgnoreCase(property)
      if ("tags".equalsIgnoreCase(property)) {
        this.includeTags = true;
      } else if ("metadata".equalsIgnoreCase(property)) {
        this.includeOSMMetadata = true;
      } else if (oldUnclippedParameter) {
        this.unclipped = true;
      } else {
        throw new BadRequestException(ExceptionMessages.PROPERTIES_PARAM);
      }
    }
```

tyrasd:
aah, regarding my comment above https://github.com/GIScience/ohsome-api/pull/29#discussion_r473817843, I realize now that it could result in double-negatives sometimes: `unclipped=false` could be a bit confusing. Maybe turn the whole thing around, and just call this parameter `clipGeometries` (or so)?!

FabiKo117:
> aah, regarding my comment above [#29 (comment)](https://github.com/GIScience/ohsome-api/pull/29#discussion_r473817843), I realize now that it could result in double-negatives sometimes: `unclipped=false` could be a bit confusing. Maybe turn the whole thing around, and just call this parameter `clipGeometries` (or so)?!

that's a good point yes.. then it would be better to call it "clipGeometries"

PR #33, opened 2020-08-24
Adapting GroupBy Requests in CSV format

### LINKED ISSUE #25: GroupBy Request with format CSV fails if result is empty
Minimal example to trigger bug (thx @tyrasd ):

```bash
curl "https://api.ohsome.org/v1/elements/count/groupBy/tag?bboxes=8.625%2C49.3711%2C8.7334%2C49.4397&filter=building%3Dnonono&format=csv&groupByKey=building"
```

Problem probably here, accessing an item in an empty array:
https://github.com/GIScience/ohsome-api/blob/1.0.0/src/main/java/org/heigit/ohsome/ohsomeapi/executor/ExecutionUtils.java#L294

bonaparten:
Adapts response in csv format if result is empty. Closes #25.


FabiKo117 (approved):
works 👍

Issue #23, opened 2020-07-30
Enhancing Data Extraction Endpoints


KowatschFab:
Through the recent discussions in https://github.com/GIScience/ohsome-api/pull/18, we've defined a need for a more structured way to represent the needed functionality to request modified data. We see the need for additional endpoints listed below. We've chosen to add these features to a new `/contributions/` endpoint as this fits well with future directions of upcoming features.

- [x] `/contributions/geometry/` for the modifications as GeoJSON
- [x] `/contributions/latest/geometry/` for the "latest" changes only, as GeoJSON

Here, `geometry` could be extended to also allow `centroid` or `bbox`.

Perhaps in the future the following endpoints could be added as well (specifics are to be defined still):

- [x] `/contributions/count` for modification counts as JSON -> #121 
- [ ] `/contributions/count/groupBy/contributionType` for modification counts as JSON grouped by contribution type (creation/modification/deletion)
- [ ] `/contributions/` for the modifications as annotated JSON (like OsmChange XML – open question: how does the format look like exactly?)
- [x] `/contributions/latest/` for the "latest" modifications as annotated JSON (like OsmChange XML)

FabiKo117:
I've started to implement the /contributions endpoint and stumbled upon some thoughts during that.

So what is the difference between the response of /fullHistory vs. the response here of /contributions? As I see it, it's only in the properties of each response feature, because you would get the same features in both endpoints for the same parameters. Currently the /fullHistory endpoint also gives one feature for each modification, it just does not add the contribution type and the contribution time to it.

So I thought it could make sense here to integrate now /fullHistory under the /contributions name and think about what kind of properties each feature must have (and could have depending on what the user wants to know).

tyrasd:
> So what is the difference between the response of /fullHistory vs. the response here of /contributions? 

To me the main difference is that `/elementsFullHistory/` not only returns the changes in the data, but also the unchanged data: i.e. if you request only a short time frame, then you will get the state as it was as the start of the time frame, and you get all changes to end up with the state at the end of the requested time frame. On the other hand the `/contributions` endpoint does not return any elements which are not touched in the time frame.

> So I thought it could make sense here to integrate now /fullHistory under the /contributions

[...]

PR #36, opened 2020-08-26
improved osmCreateFeature tag handling

rtroilo:
The original version pushes all OSHDBTags into the features properties map  and  they need then to be  extracted again for converting to OSMTags.

This pull request makes this extraction easier by not  pushing each tag individual into the features properties, but push that as a whole into a `@tags` properties. This makes the handling of those tags easier.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## current master
 
+* data extraction: escape tags with `@` characters ([#40](https://github.com/GIScience/ohsome-api/issues/40))
+
 ## 1.2.3
 
 ### Bug Fixes
--- a/docs/response-parameters.rst
+++ b/docs/response-parameters.rst
@@ -11,6 +11,13 @@ Description of parameters that are present in every response.
 * ``attribution`` - copyrights and attribution
 * ``apiVersion`` - version of the ohsome API
 
+OSM Tags
+--------
+
+When requested, the result will contain OSM elements' tags as individual GeoJSON properties.
+
+.. note:: Any OSM tag with a key starting with the special `@` character (e.g.: `@osmId`), will be modified through the addition of another `@` at the start (the example would be changed to `@@osmId`).
+
 Aggregation Parameters
 ----------------------
 
--- a/src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/DataRequestExecutor.java
+++ b/src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/DataRequestExecutor.java
@@ -83,12 +83,7 @@ public class DataRequestExecutor extends RequestExecutor {
     }
     TagTranslator tt = DbConnData.tagTranslator;
     String[] keys = requestParameters.getKeys();
-    int[] keysInt = new int[keys.length];
-    if (keys.length != 0) {
-      for (int i = 0; i < keys.length; i++) {
-        keysInt[i] = tt.getOSHDBTagKeyOf(keys[i]).toInt();
-      }
-    }
+    final Set<Integer> keysInt = ExecutionUtils.keysToKeysInt(keys, tt);
     ExecutionUtils exeUtils = new ExecutionUtils(processingData);
     inputProcessor.processPropertiesParam();
     inputProcessor.processIsUnclippedParam();
--- a/src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/ElementsRequestExecutor.java
+++ b/src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/ElementsRequestExecutor.java
@@ -10,6 +10,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Set;
 import java.util.SortedMap;
 import java.util.TreeMap;
 import java.util.stream.Collectors;
@@ -108,12 +109,7 @@ public class ElementsRequestExecutor {
     RequestParameters requestParameters = processingData.getRequestParameters();
     TagTranslator tt = DbConnData.tagTranslator;
     String[] keys = requestParameters.getKeys();
-    int[] keysInt = new int[keys.length];
-    if (keys.length != 0) {
-      for (int i = 0; i < keys.length; i++) {
-        keysInt[i] = tt.getOSHDBTagKeyOf(keys[i]).toInt();
-      }
-    }
+    final Set<Integer> keysInt = ExecutionUtils.keysToKeysInt(keys, tt);
     final MapReducer<Feature> preResult;
     ExecutionUtils exeUtils = new ExecutionUtils(processingData);
     preResult = mapRed.map(snapshot -> {
--- a/src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/ExecutionUtils.java
+++ b/src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/ExecutionUtils.java
@@ -8,6 +8,8 @@ import com.fasterxml.jackson.core.util.DefaultIndenter;
 import com.fasterxml.jackson.core.util.DefaultPrettyPrinter;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.SerializationFeature;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Streams;
 import com.opencsv.CSVWriter;
 import com.zaxxer.hikari.HikariDataSource;
 import java.io.ByteArrayOutputStream;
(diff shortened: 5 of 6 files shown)


src/main/java/org/heigit/ohsome/ohsomeapi/executor/ElementsRequestExecutor.java
```diff
@@ -115,11 +116,14 @@ public static void extract(ElementsGeometry elemGeom, HttpServletRequest servlet
...
     String[] keys = requestParameters.getKeys();
-    int[] keysInt = new int[keys.length];
+    final Set<Integer> keysInt;
     if (keys.length != 0) {
+      keysInt = new HashSet<>(keys.length);
       for (int i = 0; i < keys.length; i++) {
-        keysInt[i] = tt.getOSHDBTagKeyOf(keys[i]).toInt();
+        keysInt.add(tt.getOSHDBTagKeyOf(keys[i]).toInt());
       }
+    }else {
```
tyrasd, line 125: missing space between `}` and `else` :wink:

src/main/java/org/heigit/ohsome/ohsomeapi/executor/ExecutionUtils.java
```diff
@@ -92,6 +82,17 @@
 import org.locationtech.jts.geom.Polygonal;
 import org.locationtech.jts.geom.Puntal;
 import org.wololo.jts2geojson.GeoJSONWriter;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
```
tyrasd, line 85: this seems wrong, isn't the order of imports supposed to be alphabetical?

src/main/java/org/heigit/ohsome/ohsomeapi/executor/ExecutionUtils.java
```diff
@@ -970,9 +970,11 @@ private void writeStreamResponse(ThreadLocal<JsonGenerator> outputJsonGen,
       threadPool.submit(() -> stream.parallel().map(data -> {
         Map<String, Object> props = data.getProperties();
         OSHDBTag[] tags = (OSHDBTag[]) props.remove("@tags");
-        for (OSHDBTag tag : tags) {
-           OSMTag osmTag = tts.get().getOSMTagOf(tag);
-           props.put(osmTag.getKey(), osmTag.getValue());
+        if(tags !=  null) {
```
tyrasd, line 973: missing space here as well :stuck_out_tongue:

tyrasd: and duplicate spaces before `null`

src/main/java/org/heigit/ohsome/ohsomeapi/executor/ExecutionUtils.java
```diff
@@ -984,21 +974,16 @@ private void writeStreamResponse(ThreadLocal<JsonGenerator> outputJsonGen,
...
-          if (key.charAt(0) != '@') {
-            keysToDelete.add(key);
-            tagsToAdd.add(tts.get().getOSMTagOf((OSHDBTag) tag.getValue()));
+        Map<String, Object> props = data.getProperties();
+        OSHDBTag[] tags = (OSHDBTag[]) props.remove("@tags");
+        if (tags != null) {
+          for (OSHDBTag tag : tags) {
+             OSMTag osmTag = tts.get().getOSMTagOf(tag);
+             String key = osmTag.getKey();
+             props.put((key.startsWith("@"))?"@"+key:key, osmTag.getValue());
```
tyrasd, line 983: sorry to nag again about whitespace, but please take a look at the warnings from _checkstyle_, e.g. https://jenkins.ohsome.org/job/ohsome-api/view/change-requests/job/PR-36/8/checkstyle/new/category.-2024590691/source.62e6791e-ae76-48cc-b094-b0a6a77a47d3/#983 and https://jenkins.ohsome.org/job/ohsome-api/view/change-requests/job/PR-36/8/checkstyle/new/category.-641434199/source.93e95eb6-0bf9-4df4-b1ae-6beb57ee9e56/#983

src/main/java/org/heigit/ohsome/ohsomeapi/executor/ExecutionUtils.java
```diff
@@ -391,28 +392,17 @@ public void writeCsvResponse(Result[] resultSet, HttpServletResponse servletResp
...
-        for (int key : keysInt) {
-          if (tagKeyId == key) {
-            properties.put(String.valueOf(tagKeyId), new OSHDBTag(tagKeyId, tagValueId));
-            break;
-          }
-        }
-      }
+      properties.put("@tags", Iterables.toArray(entity.getTags(), OSHDBTag.class));
+    } else if (!keysInt.isEmpty()) {
+      Iterable<OSHDBTag> filtered = Iterables.filter(entity.getTags(), t -> keysInt.contains(t.getKey()));
```
FabiKo117, line 404: This line here is longer than the maximal allowed line length defined by our checkstyle file.

src/main/java/org/heigit/ohsome/ohsomeapi/executor/ElementsRequestExecutor.java
```diff
@@ -150,6 +146,19 @@ public static void extract(ElementsGeometry elemGeom, HttpServletRequest servlet
     }
   }
 
+  private static Set<Integer> keysToKeysInt(String[] keys, TagTranslator tt) {
```
FabiKo117, line 149: The ElementsRequestExecutor class does not have any private helper methods in it. I'd prefer if it would stay that way and you maybe put this into the helper class ExecutionUtils.java :)

I'm open though for discussion to make a bigger refactoring and structure the methods differently at some point.

FabiKo117 (changes requested):
as explained in the comments

tyrasd:
@rtroilo can you please sign your commits using a gpg key?

tyrasd:
I guess some documentation update would be needed in order to fully address https://github.com/GIScience/ohsome-api/issues/40#issuecomment-683611726

@FabiKo117 can you give me a hint where this should be mentioned?

FabiKo117:
> I guess some documentation update would be needed in order to fully address [#40 (comment)](https://github.com/GIScience/ohsome-api/issues/40#issuecomment-683611726)
> 
> @FabiKo117 can you give me a hint where this should be mentioned?

We could add a note about that after the introduction sentence on top of the response parameters [docs page](https://docs.ohsome.org/ohsome-api/stable/response-parameters.html). Something like:

"Please note: if any of the responded OSM elements has a tag with the same key as one of our custom parameters (e.g.: `@osmId`), it is modified through the addition of another @ at the start changing it to `@@osmId`."

Side-note: This functionality is not implemented yet though in that PR as far as I can see, right?

tyrasd:
> Side-note: This functionality is not implemented yet though in that PR as far as I can see, right?

It should be implemented here [in `ExecutionUtils.java` line 914](https://github.com/GIScience/ohsome-api/pull/36/files#diff-98a5146fb23f3bb10e7d6f33b7162013f9874cedba7d53a2843bf3235784f805R914). Are there more places where this needs to be adapted?

FabiKo117:
> > Side-note: This functionality is not implemented yet though in that PR as far as I can see, right?
> 
> It should be implemented here [in `ExecutionUtils.java` line 914](https://github.com/GIScience/ohsome-api/pull/36/files#diff-98a5146fb23f3bb10e7d6f33b7162013f9874cedba7d53a2843bf3235784f805R914). Are there more places where this needs to be adapted?

This just adds one additional @ to any tag that already has a leading one, right? I was thinking of a more sophisticated approach to only apply this to tags that would have the exact same key name as one of our internally-defined response parameters. But on the other hand, just escaping it once for all OSM tags would also solve this issue in general, as we would not have to change anything in case we might add or remove additional response parameters in future. I guess we can leave it then like this yes 👍 

But the suggested sentence for the docs would then have to be adapted, can do that real quick in my former comment.

tyrasd:
> I was thinking of a more sophisticated approach to only apply this to tags that would have the exact same key name as one of our internally-defined response parameters. 

that would be a bit different from what we discussed in https://github.com/GIScience/ohsome-api/issues/23#issuecomment-682455373 and #40. 

> But on the other hand, just escaping it once for all OSM tags would also solve this issue in general, as we would not have to change anything in case we might add or remove additional response parameters in future. I guess we can leave it then like this yes

:+1: 

> But the suggested sentence for the docs would then have to be adapted, can do that real quick in my former comment.

I have slightly adapted it with d100db25f2ef1eb0e88ca4e3bb9d81d6a25d5ebb. Does it look ok like that?

FabiKo117:
> I have slightly adapted it with [d100db2](https://github.com/GIScience/ohsome-api/commit/d100db25f2ef1eb0e88ca4e3bb9d81d6a25d5ebb). Does it look ok like that?

Yes, looks fine to me 👍

tyrasd (approved):

FabiKo117 (approved):

PR #38, opened 2020-08-26
upgrade filter library to version 1.2

tyrasd:
which includes the following changes: https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/libs/ohsome-filter/-/blob/1.2.0/CHANGELOG.md#anchor-120

### DIFF
--- a/docs/filter.rst
+++ b/docs/filter.rst
@@ -33,8 +33,21 @@ Selectors
     |                        | | have any tag with the given key  |                        |
     |                        | |                                  |                        |
     +------------------------+------------------------------------+------------------------+
+    | ``key in (value list)``| | matches all entities which do    | ``highway in           |
+    |                        | | have any tag with the given key  | (residential,          |
+    |                        | | and one of the given values      | living_street)``       |
+    +------------------------+------------------------------------+------------------------+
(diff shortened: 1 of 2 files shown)


FabiKo117 (changes requested):
Please also add these new features in the docs of the ohsome API under the dedicated [filter page](https://docs.ohsome.org/ohsome-api/v1/filter.html).

docs/filter.rst
```diff
@@ -124,18 +137,16 @@ Here's some useful examples for querying some OSM features:
     |                  |                                                        | | source), but is in reality |
     |                  |                                                        | | not one.                   |
     +------------------+--------------------------------------------------------+------------------------------+
-    | | highways       | | ``type:way and (highway=motorway or``                | | The list of used tags      |
```
tyrasd, line 127: this old clumsy example can now be written much more nicely with the new `key in (value list)` syntax :sunglasses:

FabiKo117: true, thanks for updating that 👍

FabiKo117 (approved):
looks good now 👍

PR #39, opened 2020-08-26
Add additional informational badges to README

joker234:
Add additional informational badges to README

Related to https://github.com/GIScience/oshdb/pull/262 and https://github.com/GIScience/badges/pull/2.

See https://github.com/GIScience/ohsome-api/tree/update-badges#ohsome-api for a preview.


joker234:
One suggestion would be, that we could move the badges to the top. What do you think?

tyrasd (approved):
looks good to me :+1: @FabiKo117: do you miss anything?

FabiKo117 (commented):
Yeah I like these badges too. I just browsed through some other GitHub projects and all of them either have the badges right next to the name in the top line, or directly under the name (like the ORS has it too).

The one right next to the name does not look that great imo, especially as we have several badges, but right under "ohsome API" would also fit.

joker234:
moved under the headline, rebased to master, can be merged

PR #45, opened 2020-09-03
updating version to 1.1.0

FabiKo117:
preparing dedicated links and version mentions for v1.1.0

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -13,7 +13,7 @@
   <name>ohsome API</name>
   <description>A public Web-RESTful-API for "ohsome" OpenStreetMap history data.</description>
   <packaging>jar</packaging>
-  <version>1.0.0-SNAPSHOT</version>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -13,7 +13,7 @@
   <name>ohsome API</name>
   <description>A public Web-RESTful-API for "ohsome" OpenStreetMap history data.</description>
   <packaging>jar</packaging>
-  <version>1.0.0-SNAPSHOT</version>
+  <version>1.1.0</version>
```
tyrasd, line 16: this we normally only change in the release branch, not in master.

FabiKo117: so to which version would it then change here?

FabiKo117: something like 1.2.0-SNAPSHOT?

tyrasd (approved):

PR #47, opened 2020-09-04
Request executor refactoring

FabiKo117:
refactoring of aggregate endpoint processing + groupBy/boundary
creating additional classes and removing unused methods
adding Java 11 specific keyword 'var' to some variables to reduce length of long variable declarations


rtroilo (commented):
We peer-coded this change together but I unified the writeCsvResponse in the AggregateRequestExecutor afterwards. So @FabiKo117  can you please check this part again, so that we can approve this together.

rtroilo (approved):

PR #49, opened 2020-09-11
docs improvements

### LINKED ISSUE #43: Improve documentation of groupByKey(s)
The following is missing:

- an example on how to specify groupByKey
- the note that when aggregating by Tag one cannot specify multiple groupByKeys with groupByValue
- the examples for groupByKey etc. are scattered around the endpoint examples but it might be a good idea to place them directly here: https://docs.ohsome.org/ohsome-api/v1/group-by.html

### LINKED ISSUE #48: document timezone of time parameter
see https://github.com/GIScience/oshdb/pull/265#discussion_r485431051. it might be good to mention in https://docs.ohsome.org/ohsome-api/v1/time.html that all timestamps supplied to the ohsome api need to be in the UTC (`Z`) time zone.

FabiKo117:
adding info about UTC timestamps to docs (fixes #48)
modifying groupBy docs page (fixes #43)

### DIFF
--- a/docs/filter.rst
+++ b/docs/filter.rst
@@ -53,7 +53,7 @@ Selectors
     |                        | | geometry of the given type       |                        |
     |                        | | (point, line, polygon, or other) |                        |
     +------------------------+------------------------------------+------------------------+
-.. [#] keep in mind that osm ids are not unique between osm types. In order to exclude only a specific object the id needs to be combined with an osm type filter e.g. (not id:10905729 and type:way) 
+.. [#] keep in mind that osm ids are not unique between osm types. In order to exclude only a specific object the id needs to be combined with an osm type filter e.g. ``type:node and id:1234``
 
 |
 
--- a/docs/filter.rst
+++ b/docs/filter.rst
@@ -41,7 +41,7 @@ Selectors
     |                        | | given osm type                   |                        |
     +------------------------+------------------------------------+------------------------+
     | ``id:osm-id``          | | matches all entities with the    | ``id:1234``            |
-    |                        | | given osm id                     |                        |
+    |                        | | given osm id [#]_                |                        |
     +------------------------+------------------------------------+------------------------+
     | ``id:(id list)``       | | matches all entities with the    | ``id:(1, 42, 1234)``   |
     |                        | | given osm ids                    |                        |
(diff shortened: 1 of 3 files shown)


docs/group-by.rst
```diff
@@ -50,10 +54,15 @@ Groups the result by the given tags that are defined through the ``groupByKey``
 
     :param aggregation: aggregation type, one of ``area``, ``count``, ``length``, ``perimeter``
     :query <params>: see query-parameters_ at /elements(aggregation) endpoint
-    :query groupByKey: OSM key(s) given as a list and combined with the ‘AND’ operator, e.g.: 'highway’, 'building’; (one groupByKey parameter must be defined)
+    :query groupByKey: OSM key e.g.: 'highway’, 'building’; no default value (one groupByKey parameter must be defined)
```
tyrasd, line 57: isn't the _one_ in _“(one groupByKey parameter must be defined)”_ superfluous and/or confusing, since there can and must be exactly one. better just say something like: _“(this parameter must be defined)”_?

FabiKo117: hmm.. yeah this could be rephrazed a bit.. your suggestion sounds fine. Let's see then what @SlowMo24 says :)

SlowMo24: I added my version in commit. I like to be more verbose. giving an example with two keys while only one can be specified seemed confusing for me

SlowMo24: one might also think about the default: no value. although it is technically correct semantically it means 'all values' ;-)

docs/filter.rst
```diff
@@ -53,6 +53,7 @@ Selectors
     |                        | | geometry of the given type       |                        |
     |                        | | (point, line, polygon, or other) |                        |
     +------------------------+------------------------------------+------------------------+
+.. [#] keep in mind that osm ids are not unique between osm types. In order to exclude only a specific object the id needs to be combined with an osm type filter e.g. (not id:10905729 and type:way)
```
tyrasd, line 56: good suggestion to put this into a footnote.
but… couldn't one make the example you give even clearer if the `not` was omitted? because I could potentially see some confusion of beginners reading this and wondering why one has to put the `not` here if one just wants to select a single element with a specific id (as in the example in the table)? I'd use the following as an example: `type:node and id:1234`

tyrasd:
thanks @SlowMo24 for the changes. i really like the clarifications in 8597c47. :+1:

docs/filter.rst
tyrasd: see 
dec25f3

SlowMo24: perfekt. that was my use case but the inverse is better for docs

SlowMo24 (approved):

PR #53, opened 2020-09-24
Towards bean refactoring

FabiKo117:

PR #54, opened 2020-09-24
improving spring usage

FabiKo117:
starting with adding more spring functionalities
using bean components to apply auto-injection in different classes

### DIFF
--- a/src/main/java/org/heigit/ohsome/ohsomeapi/Application.java
+++ b/src/main/java/org/heigit/ohsome/ohsomeapi/Application.java
@@ -1,27 +1,25 @@
 package org.heigit.ohsome.ohsomeapi;
 
-import com.zaxxer.hikari.HikariConfig;
+import com.google.common.base.Throwables;
+import com.zaxxer.hikari.HikariDataSource;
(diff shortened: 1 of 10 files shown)


FabiKo117:
This PR is quite outdated by now, but still I guess it would make sense to grab it up again at some point and continue the started work. As it's more a structural change, it could also wait until after our review I guess.

bonaparten:
> This PR is quite outdated by now, but still I guess it would make sense to grab it up again at some point and continue the started work. As it's more a structural change, it could also wait until after our review I guess.

I was thinking the same. You are right. After the review, we should work on it.

FabiKo117:
> I was thinking the same. You are right. After the review, we should work on it.
> Can we close [#53](https://github.com/GIScience/ohsome-api/pull/53) or should it stay still open?
@bonaparten It should stay open, as there has most likely not everything been considered and transfered over to this new branch and PR. Is not a nice situation, but as there's a lot of code and work involved, I'd prefer to keep it to not lose the achieved progress and thoughts.

tyrasd:
closing stale branch

PR #60, opened 2020-10-13
allowing pipe sign in GET requests

### LINKED ISSUE #59: Pipe sign delimiter in boundary parameter for GET requests not working
When using the pipe sign delimiter to give a list of different boundaries, the current master returns a 400 - BadRequestException. The following message appears in the console when tested locally:
```
java.lang.IllegalArgumentException: Invalid character found in the request target. The valid characters are defined in RFC 7230 and RFC 3986
```
used URL to test:
```
https://api.ohsome.org/v1/elements/count/groupBy/boundary?bboxes=Heidelberg:8.5992,49.3567,8.7499,49.4371|Heilbronn:9.1638,49.113,9.2672,49.1766&filter=type:way and building=residential&time=2015-01-01
```

FabiKo117:
adding of TomcatConfig class which defines the pipe sign as a valid property
needed due to recent security changes (https://bugzilla.redhat.com/show_bug.cgi?id=1397484) in Tomcat
fixes #59

### DIFF
--- a/src/main/java/org/heigit/ohsome/ohsomeapi/config/TomcatConfig.java
+++ b/src/main/java/org/heigit/ohsome/ohsomeapi/config/TomcatConfig.java
@@ -1,0 +1,25 @@
+package org.heigit.ohsome.ohsomeapi.config;
+
+import org.apache.catalina.connector.Connector;
+import org.springframework.boot.web.embedded.tomcat.TomcatConnectorCustomizer;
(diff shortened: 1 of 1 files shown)


FabiKo117:
As this issue is also a topic in the current v1.1, I think we also need to fix it there and make a release 1.1.1, right? @tyrasd

bonaparten (approved):
Looks fine

PR #62, opened 2020-10-15
upgrade ohsome-filter to 1.3.0

tyrasd:
[new features](https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/libs/ohsome-filter/-/blob/master/CHANGELOG.md#anchor-130) included here are:

* new filter: geometry filters return only features with a certain range of length (in meters) or area (in m²). example: `area:(10..20)` or `length:(..123)`
* new filter: id filter allows to select by type and id in a single expression (syntax: `id:node/42`)
* be a bit more forgiving with extra or omitted whitespace around parentheses

todo:
* [x] update ohsome-api documentation


FabiKo117 (approved):
Looks good 👍

PR #73, opened 2020-11-10
Extend info on timeout parameter

### LINKED ISSUE #66: Extend description of timeout parameter
The description in api.ohsome.org

`timeout – custom timeout to limit the processing time in seconds; default: empty`

is misleading. One the one hand a server side max-timeout applies (that might be documented at `/metadata` ) on the other the default is not 'empty' but 'some number'.

bonaparten:
Resolves #66

### DIFF
--- a/docs/response-parameters.rst
+++ b/docs/response-parameters.rst
@@ -54,7 +54,7 @@ Metadata
 
 * ``timeout`` - limit of the processing time in seconds
 * ``spatialExtent`` - spatial boundary of the OSM data in the underlying OSHDB
-* ``temporalExtent`` - temporal boundary of the OSM data in the underlying OSHDB
+* ``temporalExtent`` - timeframe of the OSM data in the underlying OSHDB
 * ``fromTimestamp`` - temporal starting point
 * ``toTimestamp`` - temporal ending point
 * ``replicationSequenceNumber`` - precise state of the OSM data contained in the underlying OSHDB, expressed as the id of the last applied (hourly) diff file from planet.openstreetmap.org
--- a/docs/response-parameters.rst
+++ b/docs/response-parameters.rst
@@ -43,7 +43,6 @@ Descriptions of the custom response parameters that are marked with a leading ``
 * ``@version`` - version of the feature
 * ``@changesetId``- id assigned to a changeset
 * ``@osmType``- type (node/way/relation) of OSM feature
-
 * ``@timestamp`` - indicates when this contribution occurred
 * ``@creation``	- contribution type; indicates if this feature is newly created (true); cannot occur in combination with other contribution types
 * ``@geometryChange`` - contribution type; indicates if the geometry of this feature has changed (true); can occur in combination with @tagChange
@@ -54,11 +53,8 @@ Metadata
 --------
 
 * ``timeout`` - limit of the processing time in seconds
-* ``extractRegion`` - considered region
-* ``spatialExtent`` - spatial respresentation of the region
-* ``type`` - indicates the type of OSM elements
-* ``coordinates`` - coordinates of the OSM elements
-* ``temporalExtent`` - represents the temporal range
-* ``fromTimestamp`` - indicates the temporal starting point
-* ``toTimestamp`` - indicates the temporal ending point
-* ``replicationSequenceNumber`` - sequence number of the change/state file
+* ``spatialExtent`` - spatial boundary of the OSM data in the underlying OSHDB
+* ``temporalExtent`` - temporal boundary of the OSM data in the underlying OSHDB
+* ``fromTimestamp`` - temporal starting point
+* ``toTimestamp`` - temporal ending point
+* ``replicationSequenceNumber`` - hourly id of the diff files from planet.openstreetmap.org
--- a/docs/response-parameters.rst
+++ b/docs/response-parameters.rst
@@ -19,10 +19,10 @@ Extraction Parameters
 
 Descriptions of the custom response parameters that are marked with a leading ``@``.
 
-* ``@osmId`` - id of the OSM feature with its type (node/way/relation)
-* ``@version`` - version of the feature
-* ``@changesetId``- id assigned to a changeset
-* ``@osmType``- type (node/way/relation) of OSM feature
+* ``@osmId`` - id of the OSM element, including its type (e.g. node/1)
+* ``@version`` - version number of the OSM element
+* ``@changesetId`` - id of the OSM changeset which last increased the version of this OSM element
+* ``@osmType`` - type of the OSM element (NODE, WAY or RELATION)
 
 **specific for /elements**
 
@@ -39,10 +39,10 @@ Contribution Parameters
 
 Descriptions of the custom response parameters that are marked with a leading ``@``.
 
-* ``@osmId`` - id of the OSM feature with its type (node/way/relation)
-* ``@version`` - version of the feature
-* ``@changesetId``- id assigned to a changeset
-* ``@osmType``- type (node/way/relation) of OSM feature
+* ``@osmId`` - id of the OSM element, including its type (e.g. node/1)
+* ``@version`` - version of the OSM element
+* ``@changesetId`` - id of the OSM changeset where the contribution was performed
+* ``@osmType`` - type of the OSM element (NODE, WAY or RELATION)
 * ``@timestamp`` - indicates when this contribution occurred
 * ``@creation``	- contribution type; indicates if this feature is newly created (true); cannot occur in combination with other contribution types
 * ``@geometryChange`` - contribution type; indicates if the geometry of this feature has changed (true); can occur in combination with @tagChange
@@ -57,4 +57,4 @@ Metadata
 * ``temporalExtent`` - temporal boundary of the OSM data in the underlying OSHDB
 * ``fromTimestamp`` - temporal starting point
 * ``toTimestamp`` - temporal ending point
-* ``replicationSequenceNumber`` - id of the hourly diff files from planet.openstreetmap.org
+* ``replicationSequenceNumber`` - precise state of the OSM data contained in the underlying OSHDB, expressed as the id of the last applied (hourly) diff file from planet.openstreetmap.org
--- a/docs/response-parameters.rst
+++ b/docs/response-parameters.rst
@@ -1,20 +1,28 @@
 Response Parameters
 ===================
 
-Descriptions of the custom response parameters that are marked with a leading ``@`` for the 
-data extraction and contributions endpoints.
+Description of the response parameters.
 
 General Parameters
 ------------------
+
+Description of parameters that are present in every response.
+
+* ``attribution`` - copyrights and attribution
+* ``apiVersion`` - version of the ohsome API
+
+Aggregation Parameters
+----------------------
+
+Extraction Parameters
+---------------------
+
+Descriptions of the custom response parameters that are marked with a leading ``@``.
 
 * ``@osmId`` - id of the OSM feature with its type (node/way/relation)
 * ``@version`` - version of the feature
 * ``@changesetId``- id assigned to a changeset
 * ``@osmType``- type (node/way/relation) of OSM feature
-
-
-Extraction Parameters
----------------------
 
 **specific for /elements**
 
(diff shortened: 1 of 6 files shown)


FabiKo117 (changes requested):
The change of the description seems good, but the /metadata request does not return the mentioned information yet :)
This still needs to be implemented as well in this PR.

src/main/java/org/heigit/ohsome/ohsomeapi/oshdb/ExtractMetadata.java
```diff
@@ -13,6 +14,7 @@
   public static Geometry dataPoly = null;
   public static JsonNode dataPolyJson = null;
   public static int replicationSequenceNumber;
+  public static double timeOutInSeconds = ProcessingData.getTimeout();
```
FabiKo117, line 17: The parameter is called "timeout", the method in the class ProcessingData.java is called getTimeout(), but here you name the variable "timeOut..." with a big "O". This does not match with the current writing of the word "timeout" in the other parts of the code.

FabiKo117 (changes requested):
see inline comment

src/main/java/org/heigit/ohsome/ohsomeapi/output/metadataresponse/MetadataResponse.java
```diff
@@ -19,10 +19,12 @@
 
   @ApiModelProperty(notes = "License and copyright info", required = true, position = 0)
   private Attribution attribution;
-  @ApiModelProperty(notes = "Version of this api", required = true, position = 0)
+  @ApiModelProperty(notes = "Version of this api", required = true, position = 1)
   private String apiVersion;
+  @ApiModelProperty(notes = "Maximal response timeout in seconds", required = true, position = 2)
+  private double timeoutInSeconds;
```
tyrasd, line 25: would it make sense to call this just `timeout` to be consistent with the naming of the query parameter: https://github.com/GIScience/ohsome-api/pull/73/files#diff-e18e0ca2c1587af2fb3550206da0889f4297c04bf4e3a1d1ee09ca46041a3c3dR25

FabiKo117: We don't have a description there in the /metadata response like we have it in other responses to describe in which unit the represented value is given, so that's why I guess it makes sense to include it here in the name.

We are also not always consistent with the unit, as in the startup parameter we use miliseconds and in the query parameter that you can set via the request we use seconds.

FabiKo117 (approved):

src/main/java/org/heigit/ohsome/ohsomeapi/output/metadataresponse/MetadataResponse.java
tyrasd: well, there is https://docs.ohsome.org/ohsome-api/v1/endpoints.html#metadata-endpoint for documentation.

for "the other responses" you mean the `showMetadata=true` output… that's true. but still one has to look at the documentation in order to know what unit the `timeout` parameter's values have to be supplied in.

A point could also be made that the `timestamp` fields in the results are not "described" (via `showMetadata`), nor is the field called `timestampAsIso8601TimeString`. :thinking:

FabiKo117: Ah true, https://docs.ohsome.org/ohsome-api/v1/endpoints.html#metadata-endpoint this would actually also have to be adapted including the new parameter @bonaparten.
And yes, we can also add a output parameter description part then to https://docs.ohsome.org/ohsome-api/v1/response-parameters.html for the /metadata response where we can then describe the unit. If we have that, it's definitely fine to call it just "timeout".

For the time-related parameters we have https://docs.ohsome.org/ohsome-api/v1/time.html. So yeah we should just collect the relevant info in the docs, then it should be fine.

FabiKo117 (changes requested):
see https://github.com/GIScience/ohsome-api/pull/73#discussion_r530460950

FabiKo117 (approved):

src/main/java/org/heigit/ohsome/ohsomeapi/output/metadataresponse/MetadataResponse.java
```diff
@@ -19,10 +19,12 @@
 
   @ApiModelProperty(notes = "License and copyright info", required = true, position = 0)
   private Attribution attribution;
-  @ApiModelProperty(notes = "Version of this api", required = true, position = 0)
+  @ApiModelProperty(notes = "Version of this api", required = true, position = 1)
```
tyrasd, line 22: puh, perhaps this could be slightly confusing, because as a user I would expect `v1` to be the “_Version of this API_”…

```suggestion
  @ApiModelProperty(notes = "Software version running this API", required = true, position = 1)
```

docs/response-parameters.rst
```diff
@@ -29,8 +37,24 @@ Extraction Parameters
 Contribution Parameters
 -----------------------
 
+Descriptions of the custom response parameters that are marked with a leading ``@``.
+
+* ``@osmId`` - id of the OSM feature with its type (node/way/relation)
```
tyrasd, line 42:
```suggestion
* ``@osmId`` - id of the OSM element, including its type (e.g. node/1)
```

docs/response-parameters.rst
```diff
@@ -29,8 +37,24 @@ Extraction Parameters
...
+* ``@version`` - version of the feature
```
tyrasd, line 43:
```suggestion
* ``@version`` - version of the OSM element
```

docs/response-parameters.rst
```diff
@@ -29,8 +37,24 @@ Extraction Parameters
...
+* ``@changesetId``- id assigned to a changeset
```
tyrasd, line 44:
```suggestion
* ``@changesetId`` - id of the OSM changeset where the contribution was performed
```

docs/response-parameters.rst
```diff
@@ -29,8 +37,24 @@ Extraction Parameters
...
+* ``@osmType``- type (node/way/relation) of OSM feature
```
tyrasd, line 45:
```suggestion
* ``@osmType`` - type of the OSM element (NODE, WAY or RELATION)
```

docs/response-parameters.rst
```diff
@@ -1,21 +1,29 @@
...
 
+Aggregation Parameters
+----------------------
 
 Extraction Parameters
 ---------------------
 
+Descriptions of the custom response parameters that are marked with a leading ``@``.
+
+* ``@osmId`` - id of the OSM feature with its type (node/way/relation)
```
tyrasd, line 22: IMHO we should be more consistent with wordings here. Since the API endpoints use the term _elements_, this should be more appropriate:

```suggestion
* ``@osmId`` - id of the OSM element, including its type (e.g. node/1)
```

docs/response-parameters.rst
```diff
@@ -1,21 +1,29 @@
...
+* ``@version`` - version of the feature
```
tyrasd, line 23:
```suggestion
* ``@version`` - version number of the OSM element
```

docs/response-parameters.rst
```diff
@@ -1,21 +1,29 @@
...
+* ``@changesetId``- id assigned to a changeset
```
tyrasd, line 24: huh? not sure what was meant by this line… maybe this would be better:

```suggestion
* ``@changesetId`` - id of the OSM changeset which last increased the version of this OSM element
```

docs/response-parameters.rst
```diff
@@ -1,21 +1,29 @@
...
+* ``@osmType``- type (node/way/relation) of OSM feature
```
tyrasd, line 25:
```suggestion
* ``@osmType`` - type of the OSM element (NODE, WAY or RELATION)
```

tyrasd (changes requested):
I found a few typos and some places where the documentation could be improved. See the inline comments and suggested changes.

docs/response-parameters.rst
```diff
@@ -29,8 +37,24 @@ Extraction Parameters
...
 * ``@timestamp`` - indicates when this contribution occurred
 * ``@creation``	- contribution type; indicates if this feature is newly created (true); cannot occur in combination with other contribution types
 * ``@geometryChange`` - contribution type; indicates if the geometry of this feature has changed (true); can occur in combination with @tagChange
 * ``@tagChange``- contribution type; indicates if the tag of this feature has changed (true); can occur in combination with @geometryChange
 * ``@deletion`` - contribution type; indicates if the feature is deleted (true); cannot occur in combination with other contribution types
+
+Metadata
```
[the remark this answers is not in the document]
FabiKo117: They were left out yes as they are quite self-explanatory in my opinion. If we decide to really document every single response object in our JSON (and then also CSV) responses, then we should do it through this work package.

docs/response-parameters.rst
```diff
@@ -29,8 +37,24 @@ Extraction Parameters
...
+--------
+
+* ``timeout`` - limit of the processing time in seconds
+* ``spatialExtent`` - spatial boundary of the OSM data in the underlying OSHDB
+* ``temporalExtent`` - temporal boundary of the OSM data in the underlying OSHDB
```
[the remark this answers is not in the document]
FabiKo117: yeah could extend the description here a bit and include the two timestamps

docs/response-parameters.rst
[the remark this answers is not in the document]
tyrasd: ah ok. if it was on purpose it's fine for me :+1:

src/main/java/org/heigit/ohsome/ohsomeapi/output/metadataresponse/MetadataResponse.java
FabiKo117: Hmmm.. these docs here are quite old actually and are only shown in the swagger JSON file.
To me "v1" and "v1.2.3" are both expressions of the version of the API. The respective JSON field in every response is also just called "apiVersion".

Also the wording would be different, cause the API runs on a specific version, not the other way around.
What about `Specific version on which this API runs`? Using "specific" here in reference to the major.minor.patch pattern.

docs/response-parameters.rst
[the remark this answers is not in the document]
tyrasd: btw: they are even mentioned above (lines 11-12, see https://github.com/GIScience/ohsome-api/pull/73/files#diff-e4b9bb401722e926ee80355fe1bdd48119a1c86d223329a4cb37873c2a3a0a5dR11-R12), I totally overlooked it in my review

docs/response-parameters.rst
```diff
@@ -1,21 +1,29 @@
...
-* ``@version`` - version of the feature
-* ``@changesetId``- id assigned to a changeset
-* ``@osmType``- type (node/way/relation) of OSM feature
+Description of parameters that are present in every response.
 
+* ``attribution`` - copyrights and attribution
+* ``apiVersion`` - version of the ohsome API
+
+Aggregation Parameters
+----------------------
```
tyrasd, line 15: is this section empty on purpose?

FabiKo117: yes, as it's out of scope from this PR and should indicate that we want to include docs on the response formats of aggregation endpoints

src/main/java/org/heigit/ohsome/ohsomeapi/output/metadataresponse/MetadataResponse.java
tyrasd: > To me "v1" and "v1.2.3" are both expressions of the version of the API.

Well, strictly speaking it would be `v1` versus `1.2.3` (sans `v`). But I get the point that they are not independent.

> What about `Specific version on which this API runs`? Using "specific" here in reference to the major.minor.patch pattern.

This should work as well. Maybe it could even be shortened a bit:

```suggestion
  @ApiModelProperty(notes = "Specific version of this API", required = true, position = 1)
```

docs/response-parameters.rst
tyrasd: ok. could you please open a github issue for that (in the 1.3 milestone), so we don't forget it. :)

docs/response-parameters.rst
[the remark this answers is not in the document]
FabiKo117:
```suggestion
* ``temporalExtent`` - timeframe of the OSM data in the underlying OSHDB
```

tyrasd (approved):

PR #78, opened 2020-11-12
providing the /contributions response as a downloadable file

FabiKo117:
adds a respective header so that the response is provided as a downloadable file

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -13,7 +13,7 @@
   <name>ohsome API</name>
   <description>A public Web-RESTful-API for "ohsome" OpenStreetMap history data.</description>
   <packaging>jar</packaging>
-  <version>1.2.0-SNAPSHOT</version>
+  <version>1.3.0-SNAPSHOT</version>
 
   <properties>
     <apachecommons-csv.version>1.6</apachecommons-csv.version>
--- a/src/main/java/org/heigit/ohsome/ohsomeapi/utils/RequestUtils.java
+++ b/src/main/java/org/heigit/ohsome/ohsomeapi/utils/RequestUtils.java
@@ -46,14 +46,31 @@ public class RequestUtils {
   }
 
   /**
-   * Checks if the given request is requesting a data-extraction.
-   * 
+   * Checks if the given request is requesting a data extraction that can potentially return a
+   * bigger amount of GeoJSON data. This can either be through using the
+   * /elements/geometry|bbox|centroid, or the /elementsFullHistory endpoint.
+   *
    * @param url the url of the request to check
    * @return whether it is a data-extraction request, or not
    */
   public static boolean isDataExtraction(String url) {
     return url.contains("elementsFullHistory") || url.contains("elements/geometry")
         || url.contains("elements/centroid") || url.contains("elements/bbox");
+  }
+
+  /**
+   * Checks if the given request is requesting a contributions extraction that can potentially
+   * return a bigger amount of GeoJSON data. This can be through using the
+   * /contributions/(latest)/geometry|bbox|centroid endpoint.
+   *
+   * @param url the url of the request to check
+   * @return whether it is a contributions-extraction request, or not
+   */
+  public static boolean isContributionsExtraction(String url) {
+    return url.contains("contributions/geometry") || url.contains("contributions/centroid")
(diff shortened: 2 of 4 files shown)


FabiKo117:
I'd suggest to also make a 1.2.2 release then after this commit, as the compression of the response is somewhat important to avoid that tools like swagger have to deal with responses that are too big.

pom.xml
```diff
@@ -13,7 +13,7 @@
   <name>ohsome API</name>
   <description>A public Web-RESTful-API for "ohsome" OpenStreetMap history data.</description>
   <packaging>jar</packaging>
-  <version>1.2.0-SNAPSHOT</version>
+  <version>1.2.1</version>
```
FabiKo117, line 16: ok.. why was the master still on 1.2.0-SNAPSHOT? :D

I assume it should change to 1.3.0-SNAPSHOT, right?

tyrasd:
```suggestion
  <version>1.3.0-SNAPSHOT</version>
```

tyrasd (changes requested):
seems good, but the `pom`'s version in master must always be a `SNAPSHOT`.

Btw: why don't we just gzip everything? I mean, for the smaller amounts of data returned by the aggregation endpoints it doesn't do much, but it also doesn't hurt also :shrug:

tyrasd:
PS: could it be that you accidentally started this branch off of the `1.2` branch? IMHO it's better to fix a bug in master first, and only then backport patches into the version branches.

FabiKo117:
> Btw: why don't we just gzip everything? I mean, for the smaller amounts of data returned by the aggregation endpoints it doesn't do much, but it also doesn't hurt also 🤷

Good question.. I think we've discussed that already in the past, right? Because there we've also decided to just encode the responses of data extraction requests and not everyone. I guess the idea was that without gzip, the data aggregation responses are also directly displayed in the browser?

> PS: could it be that you accidentally started this branch off of the 1.2 branch? IMHO it's generally better to fix bugs in master first, and only then backport patches into the version branches.

That might be the reason actually.. still the versioning in the pom in the master was not fitting, so it's also good that this got fixed now through that "mistake" :)

tyrasd:
> > Btw: why don't we just gzip everything? I mean, for the smaller amounts of data returned by the aggregation endpoints it doesn't do much, but it also doesn't hurt also shrug
> 
> Good question.. I think we've discussed that already in the past, right? Because there we've also decided to just encode the responses of data extraction requests and not everyone. I guess the idea was that without gzip, the data aggregation responses are also directly displayed in the browser?

Hmm, as far as I can see, it's already the case that all requests are gzip compressed (see the metadata request example below, but the same is also the case for `/elemet/count`).

```
$ curl -v --compressed https://api.ohsome.org/v1/metadata
[…]
> GET /v1/metadata HTTP/1.1
> Host: api.ohsome.org
> User-Agent: curl/7.68.0
> Accept: */*
> Accept-Encoding: deflate, gzip, br
[…]
< HTTP/1.1 200 
< Date: Thu, 12 Nov 2020 14:28:40 GMT
< Server: Apache/2.4.29 (Ubuntu)
[…]
< Content-Type: application/json;charset=UTF-8
< Content-Encoding: gzip
< Vary: Accept-Encoding
< Transfer-Encoding: chunked
[…]
```

FabiKo117:
Ok.. yes.. I used the wrong description for the change that was actually made. I thought it has something to do with the gzip compression, but in fact it is about setting the response header to automatically provide the response as a downloadable file:
```response.setHeader("Content-disposition", "attachment;filename=ohsome.geojson");```

So the change proposed in this PR is actually "providing the /contributions response as a downloadable file".

File: src/main/java/org/heigit/ohsome/ohsomeapi/utils/RequestUtils.java [hunk 2]
```diff
@@ -46,7 +46,9 @@ public static boolean cacheNotAllowed(String url, String[] timeParameter) {
   }
 
   /**
-   * Checks if the given request is requesting a data-extraction.
+   * Checks if the given request is requesting a data extraction that can potentially return a
```
tyrasd, line 49: Strange, there was a trailing space at the end of this line (and a few more). I remove them now.

tyrasd (approved):
:+1: (btw: I've added a changelog entry for this one as well.)

PR #80, opened 2020-11-17
fix a regression in v1.2.0

tyrasd:
see https://github.com/GIScience/ohsome-api/pull/57#issuecomment-728838052


joker234 (approved):
Some code styling gets reverted as well, this has to be re-done in the future.

Looks good to me.

tyrasd:
> Some code styling gets reverted as well, this has to be re-done in the future.

yeah, I noticed that as well. IMHO those changes should have been in a separate commit in the first place. :shrug:

PR #82, opened 2020-11-17
checkstyle: break some too long lines and add newline after static imports

tyrasd:
Re-does the reverted code style fixes from #80 (see https://github.com/GIScience/ohsome-api/pull/80#pullrequestreview-532232325) and some more remaining code style issues in those files.

### DIFF
--- a/src/main/java/org/heigit/ohsome/ohsomeapi/executor/ExecutionUtils.java
+++ b/src/main/java/org/heigit/ohsome/ohsomeapi/executor/ExecutionUtils.java
@@ -117,7 +117,8 @@ public class ExecutionUtils {
    * Applies a filter on the given MapReducer object using the given parameters. Used in
    * /ratio/groupBy/boundary requests.
    */
-  public MapAggregator<OSHDBCombinedIndex<OSHDBTimestamp, Integer>, OSMEntitySnapshot> snapshotFilter(
+  public MapAggregator<OSHDBCombinedIndex<OSHDBTimestamp, Integer>, OSMEntitySnapshot>
+      snapshotFilter(
       MapAggregator<OSHDBCombinedIndex<OSHDBTimestamp, Integer>, OSMEntitySnapshot> mapRed,
       Set<OSMType> osmTypes1, Set<OSMType> osmTypes2, Set<SimpleFeatureType> simpleFeatureTypes1,
       Set<SimpleFeatureType> simpleFeatureTypes2, Integer[] keysInt1, Integer[] keysInt2,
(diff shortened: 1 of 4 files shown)


src/main/java/org/heigit/ohsome/ohsomeapi/executor/ExecutionUtils.java
```diff
@@ -512,28 +516,30 @@ public void writeCsvResponse(Result[] resultSet, HttpServletResponse servletResp
...
   @SuppressWarnings({"unchecked"}) // intentionally suppressed as type format is valid
-  public <K extends Comparable<K> & Serializable, V extends Number> SortedMap<OSHDBCombinedIndex<OSHDBCombinedIndex<Integer, K>, OSHDBTimestamp>, V> computeNestedResult(
+  public <K extends Comparable<K> & Serializable, V extends Number>
+      SortedMap<OSHDBCombinedIndex<OSHDBCombinedIndex<Integer, K>, OSHDBTimestamp>, V>
+      computeNestedResult(
       RequestResource requestResource,
-      MapAggregator<OSHDBCombinedIndex<OSHDBCombinedIndex<Integer, K>, OSHDBTimestamp>, Geometry> preResult)
-      throws Exception {
+      MapAggregator<OSHDBCombinedIndex<OSHDBCombinedIndex<Integer, K>, OSHDBTimestamp>, Geometry>
+          preResult) throws Exception {
```
tyrasd, line 524: normally I try to avoid breaking a variable's type and name, but here I think it can't be avoided, can it?

FabiKo117: If we want to enforce being below the 100 character per line limit, then I think it can't be yeah..

But according to https://google.github.io/styleguide/javaguide.html#s4.4-column-limit the changes in the Javadoc links are exceptions and won't have to adhere to this limit.

tyrasd: > Javadoc links

I tried it and it seems like at least the `{@link` links still work fine in the generated javadoc even if they are split across multiple lines. So I guess that at least for those _obeying the column limit is_ actually _possible_. Which means that we have to break those lines if we want to stick to the style guide. Maybe at the time when the style guide was written it was not possible to split those across multiple lines, or they actually mean regular `https://example.com/…` links inside of a javadoc comment, not sure. :shrug:

tyrasd: > If we want to enforce being below the 100 character per line limit, then I think it can't be yeah..

an alternative would be to split it like this:

```java
      MapAggregator<OSHDBCombinedIndex<OSHDBCombinedIndex<Integer, K>, OSHDBTimestamp>,
          Geometry> preResult) throws Exception {
```

or even perhaps

```java
      MapAggregator<
          OSHDBCombinedIndex<OSHDBCombinedIndex<Integer, K>, OSHDBTimestamp>, Geometry
      > preResult) throws Exception {
```

or similar variants of using the `<>` brackets in any meaningful way

FabiKo117: Okey.. yeah also not sure what exactly is meant by that. But if it works, we can break the code links in the javadoc then like you suggested yes. I guess we should at least make an issue then out of this here then to not forget it and that someone can apply this then to all classes across the project.

About the longer variable names: we've started to use the keyword ```var``` in the declaration for some very long ones. Of course this does not work in method headers, but it could be that there are still some lines in the bodies, where we could shorten the line length through using that.

bonaparten (approved):
It looks fine to me

src/main/java/org/heigit/ohsome/ohsomeapi/executor/ExecutionUtils.java
bonaparten: I took a fast look at it and it seems that the only lines which are not split in the javadocs are actually {@links.

PR #86, opened 2020-12-11
upgrade OSHDB to version 0.6

tyrasd:
see https://github.com/GIScience/oshdb/blob/master/CHANGELOG.md#060

* [x] "fix" tests after upstream changes:
    > improve accuracy of built-in geometry helper functions which calculate the geodesic lengths and areas of OSM geometries
* [ ] double-check new integrated filter implementation
* [x] add to changelog


joker234:
@tyrasd you could update the ohsome parent as well ;)

tyrasd:
> @tyrasd you could update the ohsome parent as well ;)

good point, done in c5b9f56

tyrasd:
@bonaparten @FabiKo117 I quickly checked the failing assertations from https://jenkins.ohsome.org/job/ohsome-api/view/change-requests/job/PR-86/3/consoleFull and found all "new" results to be within a 0.35% margin of the "old" test-values. Which is what I would have expected from the changes in https://github.com/GIScience/oshdb/pull/193. Did I overlook some cases where the discrepancy is larger?

FabiKo117:
> @bonaparten @FabiKo117 I quickly checked the failing assertations from https://jenkins.ohsome.org/job/ohsome-api/view/change-requests/job/PR-86/3/consoleFull and found all "new" results to be within a 0.35% (relative) margin of the "old" test-values. Which is what I would have expected from the changes in [GIScience/oshdb#193](https://github.com/GIScience/oshdb/pull/193). Did I overlook some cases where the discrepancy is larger?

@tyrasd No, it seems like this is the maximum discrepancy then. Probably I was thinking of the higher differences in bigger values, when I was referring to differenes in the lower percentage area when mentioning it in the meeting. I'd suggest to use a delta of 0.5 in the tests + adapting the current values to the more accurate ones. What do you think about that?

Another thing: all the /contribution tests are now failing for the following reason: 
```groupByEntity() must be called before any `map` or `flatMap` transformation functions have been set```
The first line in the ohsome API code that is referenced in the error message is [this one](https://github.com/GIScience/ohsome-api/blob/46cfb0f7fb311fa1e9341f9f6df7d05e05aab15b/src/main/java/org/heigit/ohsome/ohsomeapi/executor/DataRequestExecutor.java#L112).

**Comment by tyrasd** [comment]
>  I'd suggest to use a delta of 0.5 in the tests + adapting the current values to the more accurate ones. What do you think about that?

that sound good :+1: 

> Another thing: all the /contribution tests are now failing for the following reason:

oh, true. let me debug that.

**Comment by tyrasd** [comment]
> > Another thing: all the /contribution tests are now failing
> 
> oh, true. let me debug that.

Quick update: I run into the following bug while trying to fix this: https://github.com/GIScience/oshdb/issues/321 :grimacing:

**Comment by tyrasd** [comment]
> > > Another thing: all the /contribution tests are now failing
> > 
> > 
> > oh, true. let me debug that.
> 
> Quick update: I run into the following bug while trying to fix this: [GIScience/oshdb#321](https://github.com/GIScience/oshdb/issues/321) grimacing

this should now hopefully be fixed with oshdb version 0.6.1 (updated this PR with 6e66a5e34666d5975bcefab687f42deac3403ed3)

**Comment by bonaparten** [comment]
Tests are now updated with new values and delta.

File: src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/DataRequestExecutor.java [hunk 1]
```diff
@@ -52,12 +52,12 @@ public DataRequestExecutor(RequestResource requestResource, ElementsGeometry ele
    * 
    * @throws Exception thrown by
    *         {@link org.heigit.ohsome.ohsomeapi.inputprocessing.InputProcessor#processParameters()
-   *         processParameters},
-   *         {@link org.heigit.bigspatialdata.oshdb.util.time.ISODateTimeParser#parseISODateTime(String)
+   *         processParameters}, {@link
+   *         org.heigit.bigspatialdata.oshdb.util.time.IsoDateTimeParser#parseIsoDateTime(String)
    *         parseISODateTime},
```
FabiKo117, line 57:
```suggestion
   *         parseIsoDateTime},
```

src/main/lombok/org/heigit/ohsome/ohsomeapi/inputprocessing/InputProcessor.java
```diff
@@ -12,6 +12,7 @@
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.Supplier;
```
FabiKo117, line 15: I don't see where this import statement is used. There are also 10 other import statements which are apparently not used anymore in this class and would have to be removed as well.

src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/ElementsRequestExecutor.java
```diff
@@ -192,10 +192,6 @@ public static void extract(RequestResource requestResource, ElementsGeometry ele
     if (processingData.isContainingSimpleFeatureTypes()) {
       mapAgg = inputProcessor.filterOnSimpleFeatures(mapAgg);
     }
-    Optional<FilterExpression> filter = processingData.getFilterExpression();
```
FabiKo117, line 195: This removal caused the import in line 12 to be obsolete, so it would have to be removed.

src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/AggregateRequestExecutor.java
```diff
@@ -352,10 +352,6 @@ private Response writeCsv(List<String[]> comments, Consumer<CSVWriter> consumer)
     if (processingData.isContainingSimpleFeatureTypes()) {
       mapAgg = inputProcessor.filterOnSimpleFeatures(mapAgg);
     }
-    Optional<FilterExpression> filter = processingData.getFilterExpression();
```
FabiKo117, line 355: Same thing here with the imports, please check again.

src/test/java/org/heigit/ohsome/ohsomeapi/controller/TestProperties.java
```diff
@@ -10,4 +10,5 @@
   public static final String JUNIT = System.getProperty("junit");
   public static final String INTEGRATION = System.getProperty("integration");
   public static final String SERVER = "http://localhost:";
+  public static final double DELTA_PERCENTAGE = 0.5 / 100; // 0.5 %
```
FabiKo117, line 13: I could not find a specific line regarding putting comments in-line vs. before the line in the Google Java style guide, but in the ohsome API it's normally written before the respective line. I found one exception for that though [here](https://github.com/GIScience/ohsome-api/blob/c9da9c78fa5834c9b530211697fbebba7a3ae93f/src/main/lombok/org/heigit/ohsome/ohsomeapi/inputprocessing/InputProcessor.java#L129), but I'd prefer to keep it to that one exception :)

If you really want to write a comment here, I'd suggest to change it to something like:
```suggestion
  // gives a percentage of 0.5%
  public static final double DELTA_PERCENTAGE = 0.5 / 100;
```

FabiKo117 (changes requested):
Please look at the detailled review comments. There are several unused import statements from removed code parts, make sure to check again the changed classes and organize the imports there again (think mostly in classed edited by @tyrasd).

About the tests: I did not check every single test, but individual ones without the delta and they worked fine and all tests are running through of course. Please check the review comment about commenting @bonaparten.

src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/DataExtractionTransformer.java
```diff
@@ -7,19 +7,19 @@
...
 import org.heigit.bigspatialdata.oshdb.util.time.TimestampFormatter;
-import org.heigit.ohsome.filter.FilterExpression;
 import org.heigit.ohsome.ohsomeapi.controller.rawdata.ElementsGeometry;
 import org.heigit.ohsome.ohsomeapi.inputprocessing.InputProcessingUtils;
 import org.heigit.ohsome.ohsomeapi.inputprocessing.SimpleFeatureType;
 import org.locationtech.jts.geom.Geometry;
 import org.wololo.geojson.Feature;
 
-public class FlatMapExecutor implements Serializable {
+public class DataExtractionTransformer implements Serializable {
```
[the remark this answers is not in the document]
tyrasd: what's wrong with letting the jdk automatically choose one? I mean we use tons of (serialized) lambdas all over the place, and there you can't even set the UID field, even if we wanted to. :shrug:

src/test/java/org/heigit/ohsome/ohsomeapi/controller/TestProperties.java
tyrasd: I think it is easy enough to understand that dividing by 100 returns a percentage. I will just remove this comment from the PR.

README.md
```diff
@@ -59,7 +59,7 @@ To run the tests locally, you need the following:
     * -Dport.get starts data-aggregation + metadata tests using GET requests
     * -Dport.post starts data-aggregation tests using POST requests
     * -Dport.data starts data-extraction tests using GET and POST requests
-2. [heidelberg.oshb](http://downloads.ohsome.org/v0.4/germany/baden-wuerrtemberg/heidelberg.oshdb.mv.db) file (or any other, which includes the data from Heidelberg)
+2. [heidelberg.oshb](https://downloads.ohsome.org/v0.5/europe/germany/baden-wuerttemberg/heidelberg_68900_2020-07-23.oshdb.mv.db) file (or any other, which includes the data from Heidelberg)
```
joker234, line 62:
```suggestion
2. [heidelberg.oshdb](https://downloads.ohsome.org/v0.6/europe/germany/baden-wuerttemberg/heidelberg_68900_2020-07-23.oshdb.mv.db) file (or any other, which includes the data from Heidelberg)
```

src/test/java/org/heigit/ohsome/ohsomeapi/controller/TestProperties.java
FabiKo117: I agree, it should also be obvious from the variable name and the given value. But let @bonaparten due this change as it was also him who added this part, or at least he should give his opinion on that as well.

CHANGELOG.md
```diff
@@ -1,9 +1,11 @@
 ## current master
 
+* update OSHDB to version 0.6.0 ([#86](https://github.com/GIScience/ohsome-api/pull/86))
```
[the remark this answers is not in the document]
joker234: General suggestion: Do you want to apply a similar link schema as @tyrasd introduced in https://github.com/GIScience/oshdb/pull/331, which is a little bit more tidied up if you read the CHANGELOG without GitHub or another markdown parser (e.g. if you read the source from https://github.com/GIScience/ohsome-api/archive/1.2.3.zip or other sources).

Maybe it's better to create a new PR for this.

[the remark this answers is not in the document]
tyrasd: I agree that we should mention that returned length/area/perimeter values are more precise now!

src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/DataExtractionTransformer.java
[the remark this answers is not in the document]
FabiKo117: I don't have a strong opinion on whether we need to define one, or go for the automatic choose option. What I want to point out is that we partially have a custom one and partially not and that's what should be avoided in my opinion (unless it makes sense to set a custom one for specific classes and for others not?!?).

So alternatively we can also remove all custom UIDs and let it always be set automatically. That should be handled then in another PR of course.

CHANGELOG.md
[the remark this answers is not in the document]
FabiKo117: @joker234 Yes, we can go for that here in the ohsome API as well of course. Also looks a bit cleaner when all the links are together.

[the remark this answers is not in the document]
joker234: > @joker234 Yes, we can go for that here in the ohsome API as well of course. Also looks a bit cleaner when all the links are together.

Okay, I can take care of that, when this PR is merged. Then I would keep it like that in this PR.

src/test/java/org/heigit/ohsome/ohsomeapi/controller/TestProperties.java
bonaparten: No problem with deleting the comment.

src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/DataExtractionTransformer.java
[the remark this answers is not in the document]
tyrasd: let's discuss this in a separate ticket :+1:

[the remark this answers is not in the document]
FabiKo117: done https://github.com/GIScience/ohsome-api/issues/92

src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/ElementsRequestExecutor.java
FabiKo117: The obsolete import in this class is still in there.

src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/UsersRequestExecutor.java
```diff
@@ -315,10 +315,6 @@ public static Response countGroupByKey(HttpServletRequest servletRequest,
     if (processingData.isContainingSimpleFeatureTypes()) {
       mapAgg = inputProcessor.filterOnSimpleFeatures(mapAgg);
     }
-    Optional<FilterExpression> filter = processingData.getFilterExpression();
```
FabiKo117, line 318: Same applies here: still obsolete imports in the class.

src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/AggregateRequestExecutor.java
FabiKo117, line 355: And here: obsolete imports in this class.

FabiKo117 (changes requested):
See inline comments.

src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/ElementsRequestExecutor.java
tyrasd: oops, overlooked that one. sorry.

joker234:
> Looks good now. Marking the test values as 'final' also makes sense here.

I checked 3d3ec09b43e074e3bcf3a6e5877b0e7099e54b81 against a test cluster with oshdb version 0.6.1 using our usual tools and couldn't find anything suspicious.

PR #94, opened 2021-01-15
minor styling fixes in CHANGELOG and docs

joker234:


CHANGELOG.md
```diff
@@ -54,12 +54,12 @@
...
 ### Bug Fixes
 
-* fixing not-encoded pipe sign in GET requests ([#60]) ([#61])
+* fixing not-encoded pipe sign in GET requests ([#60], [#61])
 
 ### Other Changes
 
 * update to oshdb version 0.5.10 (including an Ignite update to 2.9.0) ([#67])
-* diverse code refactoring improvements, e.g. ([#46]), ([#47]), ([#51]), ([#57])
+* diverse code refactoring improvements (e.g. [#46], [#47], [#51], [#57])
```
FabiKo117, line 62: about the "e.g." it's sometimes written as "e.g.:", sometimes "e.g." and sometimes missing when listing more than one issue/pr

I guess we could just remove it in general and just list each issue/pr through the linked number.

FabiKo117 (changes requested):
see comment

CHANGELOG.md
joker234: The difference between “e.g.” being present or not would be that the list of commit/PR IDs is complete or not. If we remove the “e.g.”s, I would assume, as reader, that the list contains all code refactoring improvements (in this example).

joker234: Fixed different “e.g.”s, but not removed them (yet): https://github.com/GIScience/ohsome-api/compare/26bfa4cb250a0ac8672b5c49f75ce903af33cdf4..227476f9a3877184a67e7445e89ac6dd763cf8e5

FabiKo117: True.. but would the reader then take the effort to look manually for further refactoring improvements (which are mostly hidden in other feature commits)? Probably not.. :) but I get the point, that the message would then be a different one. We should at least have a coherent writing of "e.g.:" and can then leave it as you say yeah.

FabiKo117 (approved):
looks good 👍 

good catch on the spelling mistake + formatting improvements in the docs

PR #96, opened 2021-01-15
Fix OSHDB version

joker234:
correct is oshdb version 0.6.1, not 0.6.0 in release 1.3.0


FabiKo117 (approved):
thx for that.. 6 eyes overlooked it somehow before :D

PR #97, opened 2021-01-19
make sure groupByEntity is called before filters are set

tyrasd:
fixes a bug in the implementation of the oshdb upgrade (to version 0.6) in #86


joker234 (approved):
Resolves the addressed bug. Checked with check-ohsome-api and ohsome-api-outputcomparer.
@FabiKo117, do you want to have a look too?

FabiKo117:
> Resolves the addressed bug. Checked with check-ohsome-api and ohsome-api-outputcomparer.
> @FabiKo117, do you want to have a look too?

I was following your discussions and it seems like you resolved it, so I don't have to.
We should continue to discuss though how we can avoid that in future releases.

FabiKo117:
I can run the affected data-extraction tests then once I've updated them for https://github.com/GIScience/ohsome-api/pull/100.

FabiKo117 (approved):
Tested with updated tests from https://github.com/GIScience/ohsome-api/pull/100 and it works 👍

PR #102, opened 2021-01-21
fix filter parameter not set when using groupBy/boundary requests

### LINKED ISSUE #99: Filter not set when using groupBy/boundary request
The filter parameter seems to be ignored when using groupBy/boundary requests. 
Example: request returning the count for churches in a small bbox

`https://api.ohsome.org/v1/elements/count?bboxes=8.69002,49.41719,8.69188,49.41857&time=2019-01-01&showMetadata=true&filter=building=church and geometry:polygon`
returns correctly 1 count

`https://api.ohsome.org/v1/elements/count/groupBy/boundary?bboxes=8.69002,49.41719,8.69188,49.41857&time=2019-01-01&showMetadata=true&filter=building=church and geometry:polygon`
returns 139 counts

//edit @tyrasd : this is potentially a regression introduced by #86

tyrasd:
should fix #99 :crossed_fingers: 

this was indeed a regression introduced with #86 :see_no_evil:

### DIFF
--- a/src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/AggregateRequestExecutor.java
+++ b/src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/AggregateRequestExecutor.java
@@ -10,6 +10,7 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Optional;
 import java.util.SortedMap;
 import java.util.function.Consumer;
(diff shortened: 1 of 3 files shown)


FabiKo117:
Nice, thx! Will check it with my updated test classes from the other branch and then come back to you here.

FabiKo117 (approved):
Checked the code and tested it with updated tests in [this branch](https://github.com/GIScience/ohsome-api/tree/update-tests-with-filter) and it runs through 👍

PR #104, opened 2021-01-28
upgrade OSHDB to 0.6.3, release as 1.3.2

tyrasd:
see https://github.com/GIScience/oshdb/pull/343 for the fix in this patch


joker234 (approved):
worked for me

PR #112, opened 2021-02-04
drop unused constants, fix syntax error in filter

### LINKED ISSUE #111: Quotation marks in example values of the field `filter` cause a syntax error in Swagger UI
In Swagger UI, in the endpoint `/elements/count/ratio` quotation marks in the example value of the field `filter2` cause a syntax error.

We should also take a look at the example values in other endpoints.

tyrasd:
### Description
Fixes a syntax error in one of the default filter examples in swagger.

### Corresponding issue
Closes #111

### Checklist
- [x] I have made corresponding changes to the [documentation](https://github.com/GIScience/ohsome-api/tree/master/docs)~~
- [x] I have updated the [CHANGELOG.md](https://github.com/GIScience/oshdb/blob/master/CHANGELOG.md)

### DIFF
--- a/src/main/lombok/org/heigit/ohsome/ohsomeapi/controller/DefaultSwaggerParameters.java
+++ b/src/main/lombok/org/heigit/ohsome/ohsomeapi/controller/DefaultSwaggerParameters.java
@@ -5,13 +5,11 @@ public class DefaultSwaggerParameters {
 
   public static final String BBOX = "8.625,49.3711,8.7334,49.4397";
   public static final String HIGHWAY_KEY = "highway";
-  public static final String BUILDING_KEY = "building";
-  public static final String HOUSENUMBER_KEY = "addr:housenumber";
   public static final String RESIDENTIAL_VALUE = "residential";
   public static final String TYPE_FILTER = "type:way";
   public static final String HIGHWAY_FILTER = "highway=residential";
   public static final String BUILDING_FILTER = "building=*";
-  public static final String HOUSENUMBER_FILTER = "type:node and \"addr:housenumber=*\"";
+  public static final String HOUSENUMBER_FILTER = "type:node and \"addr:housenumber\"=*";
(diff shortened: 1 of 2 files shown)


src/main/lombok/org/heigit/ohsome/ohsomeapi/controller/DefaultSwaggerParameters.java
```diff
@@ -6,12 +6,10 @@
   public static final String BBOX = "8.625,49.3711,8.7334,49.4397";
   public static final String HIGHWAY_KEY = "highway";
   public static final String BUILDING_KEY = "building";
-  public static final String HOUSENUMBER_KEY = "addr:housenumber";
-  public static final String RESIDENTIAL_VALUE = "residential";
   public static final String TYPE_FILTER = "type:way";
   public static final String HIGHWAY_FILTER = "highway=residential";
   public static final String BUILDING_FILTER = "building=*";
-  public static final String HOUSENUMBER_FILTER = "type:node and \"addr:housenumber=*\"";
+  public static final String HOUSENUMBER_FILTER = "type:node and \"addr:housenumber\"=*";
```
FabiKo117, line 12: Would you recommend writing the key in that way using the quotation marks?

I think that tag here solely had those marks, because me (or whoever added that in the past) thought, that they are required for this specific key when using the filter parameter. Otherwise, we can just write it without the quotation marks and follow the conventions of the docs as you described it in the comment before.

tyrasd: > Would you recommend writing the key in that way using the quotation marks?

I would normally write it without the quote marks, but there's also nothing wrong with using them.

I left them in this PR because I thought it might be a hint for new users playing around in the swagger UI so find out that one can also define tag keys with `"`s.

FabiKo117 (approved):

PR #114, opened 2021-02-04
Fix slow ratio requests

rtroilo:
Ratio request were very slow in comparison to count requests.
This PR fix this issue.

### Description
change how filters were applied, so that filters are used as early as possible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,14 +3,18 @@ Changelog
 
 ## 1.4.0-SNAPSHOT (current master)
 
-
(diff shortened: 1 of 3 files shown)


tyrasd (dismissed):

tyrasd (changes requested):
please add a short line about this improvement to the [changelog](https://github.com/GIScience/ohsome-api/blob/fix_slow_ratio_requests/CHANGELOG.md)

tyrasd (approved):

Issue #115, opened 2021-02-05
Implementing aggregation endpoint(s) for /contributions


KowatschFab:
This continues with the original idea about the /contributions endpoint defined in https://github.com/GIScience/ohsome-api/issues/23.

- [x] needs https://github.com/GIScience/ohsome-api/issues/117 to be solved

A `/contributions/count` endpoint would be useful e.g. for the last edit indicator in the OQT, where the number of edits (=contributions) is the required measure.

- [x] `/contributions/count` :arrow_right: #121 
- [x] `/contributions/count/groupBy/boundary` :arrow_right: #217
- [ ] `/contributions/count/groupBy/contributionType`
- `/contributions/count/groupBy/_additional_grouping_type(tbd)`

FabiKo117:
Parts of the here mentioned features will be implemented with #121. Therefore, I'll remove the 1.4 milestone for now while still letting the issue open, as the other endpoints will most likely follow soon.

Hagellach37:
We are moving towards ohsome API v2.0.
Hence, this topic will not be followed up on.

Feel free to raise it again if it is still relevant for the new API version.

PR #121, opened 2021-02-09
implementing /contributions/count endpoint

FabiKo117:
### Description
Implementation of new endpoint /contributions/count

### Corresponding issue
#115

### New or changed dependencies
- none

### Checklist
- [x] controller
- [x] response format
- [x] processing
- [x] swagger endpoint + example parameters
- [x] https://github.com/GIScience/ohsome-api/pull/124 needs to be merged
- [x] Add info about this feature being experimental to docs
- [x] I have updated the [CHANGELOG.md](https://github.com/GIScience/oshdb/blob/master/CHANGELOG.md)


src/main/lombok/org/heigit/ohsome/ohsomeapi/controller/dataextraction/contributions/ContributionsController.java
```diff
@@ -69,7 +69,7 @@ public void contributionsBbox(HttpServletRequest servletRequest,
         ElementsGeometry.BBOX, servletRequest, servletResponse);
     executor.extract();
   }
-
+
```
tyrasd, line 72: normally, we avoid trailing white spaces also on empty lines. :wink: 

same in line 96 below

docs/endpoints.rst
```diff
@@ -1021,9 +1021,225 @@ Show number of users editing buildings before, during and after Nepal earthquake
...
+
+**Example request**:
+
+Number of contributions to the building 'Stadthalle Heidelberg' between 2010 and 2020.
+
+  .. tabs::
+
+    .. code-tab:: bash curl (GET)
+
+       curl -X GET 'https://api.ohsome.org/v1/contributions/count?bboxes=8.699053,49.411842,8.701311,49.412893&filter=id:way/140112810&time=2010-01-01,2020-01-01
```
bonaparten, line 1039: The forgotten single quote at the end of this URL causes a warning at line 1037 creating the docs with Sphinx

bonaparten (changes requested):
I know that we should avoid boilerplate code, but let some classes of the contributions endpoints be used by classes of the users endpoints could be misleading. Here some points:

- The UsersController.java uses the ContributionsExecutor.java
- The ContributionsResult.java is called by classes of users endpoints like the UsersRequestExecutor.java
- The method "count" in the class ContributionsExecutor is now used for users/count
- The method writeCsvResponse in AggregateRequestExecutor.java uses ContributionsResults.java even if this class is supposed to be used only for /elements and  /elements/_/groupBy/boundary endpoints.

Refactoring some classes and let them "extend" other in-between classes could be a solution.

//edit: adding a get test would be probably fine.

docs/endpoints.rst
FabiKo117: Good catch.. my sphinx did not give that specific warning.. just something cryptic about line 4 in this file, but after that fix the warning went away 🤷

FabiKo117:
>     * The UsersController.java uses the ContributionsExecutor.java

This is because the `/users` endpoint also uses the contribution-view of the OSHDB, hence their processing is similar. We could think about renaming the executor classes then in a way that this is better reflected and cannot be mixed up with the name of the endpoint `/contributions`. This would be one thing that should be performed with the general refactoring including the extension of the AggregageRequestExecutor class, which is not working for every endpoint yet too.

>     * The ContributionsResult.java is called by classes of users endpoints like the UsersRequestExecutor.java

Same reason as described above.

>     * The method "count" in the class ContributionsExecutor is now used for users/count

Again same reason.

>     * The method writeCsvResponse in AggregateRequestExecutor.java uses ContributionsResults.java even if this class is supposed to be used only for /elements and  /elements/_/groupBy/boundary endpoints.

This is indeed not so nice code-quality wise. But this behavior did not get introduced with this branch. It was already before using the `/users` endpoint. But you are right here, this would also have to be refactored. Maybe we can create something like a `GenericRequestExecutor` that holds methods like `writeCsvResponse` which are used through different endpoint and processing structures?

> Refactoring some classes and let them "extend" other in-between classes could be a solution.
>
> //edit: adding a get test would be probably fine.

I don't understand what you mean with these two points. A test for what exactly?

bonaparten:
> > ```
> > * The method writeCsvResponse in AggregateRequestExecutor.java uses ContributionsResults.java even if this class is supposed to be used only for /elements and  /elements/_/groupBy/boundary endpoints.
> > ```
> 
> This is indeed not so nice code-quality wise. But this behavior did not get introduced with this branch. It was already before using the `/users` endpoint. But you are right here, this would also have to be refactored. Maybe we can create something like a `GenericRequestExecutor` that holds methods like `writeCsvResponse` which are used through different endpoint and processing structures?
> 

We have this method also in ExecutionUtils.java. Yes, this could be a solution. Could you please change the javadoc upon [AggregateRequestExecutor](https://github.com/GIScience/ohsome-api/blob/bbfbdf33066065103de7da8f7fffa0d14fc92065/src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/AggregateRequestExecutor.java#L52) adding that it is used also for the contributions endpoints?

> > Refactoring some classes and let them "extend" other in-between classes could be a solution.
> > //edit: adding a get test would be probably fine.
> 
> I don't understand what you mean with these two points. A test for what exactly?

Like you already said with the class GenericRequestExecutor, I meant to create some superclasses and interfaces where needed in order to achieve more inheritance and structure.
I saw you added some tests for POST requests, you could also add some tests for GET requests.

FabiKo117:
@bonaparten I've updated the javadoc.

> I saw you added some tests for POST requests, you could also add some tests for GET requests.

It does not matter what type of HTTP request is used in the tests, as the content of GET and POST is processed equally within the code. This was different in earlier versions of the code, but now there is no difference in the processing.

docs/endpoints.rst
```diff
@@ -1028,6 +1028,8 @@ Contributions Aggregation
 
    Get the count of the contributions provided to the OSM data. This endpoint does not support the deprecated ``types``, ``keys``, ``values`` parameters.
 
+.. note:: The /contributions/count endpoint is a new feature that is in the experimental status, meaning it is still under internal evaluation and might be subject to changes in the upcoming minor or patch releases.
```
bonaparten, line 1031: Nice!

bonaparten (approved):

src/main/lombok/org/heigit/ohsome/ohsomeapi/utils/GroupByBoundaryGeoJsonGenerator.java
```diff
@@ -3,13 +3,14 @@
...
 import org.heigit.ohsome.ohsomeapi.output.Result;
-import org.heigit.ohsome.ohsomeapi.output.contributions.UsersResult;
+import org.heigit.ohsome.ohsomeapi.output.contributions.ContributionsResult;
 import org.heigit.ohsome.ohsomeapi.output.elements.ElementsResult;
 import org.heigit.ohsome.ohsomeapi.output.groupby.GroupByObject;
 import org.heigit.ohsome.ohsomeapi.output.groupby.GroupByResult;
 import org.heigit.ohsome.ohsomeapi.output.ratio.RatioGroupByResult;
 import org.heigit.ohsome.ohsomeapi.output.ratio.RatioResult;
 
+
```
bonaparten, line 13: One line too much here

bonaparten (changes requested):

tyrasd (changes requested):
I noticed that this new endpoint still supports the old types/keys/values syntax. I'd rather not have deprecated functionality included with new endpoints. Same as already discussed in https://github.com/GIScience/ohsome-api/pull/34#pullrequestreview-498497264

bonaparten (dismissed):

docs/endpoints.rst
```diff
@@ -1021,9 +1021,227 @@ Show number of users editing buildings before, during and after Nepal earthquake
...
+	      "toTimestamp":"2020-01-01T00:00:00Z",
+	      "value":15.0
+	    }
+	  ]
+	}
 
+
+.. http:post :: /contributions/count/density
+
+   Get the density of the count of contributions divided by the total area in square-kilometers. This endpoint does not support the deprecated ``types``, ``keys``, ``values`` parameters.
```
tyrasd, line 1138:
```suggestion
   Get the density of the count of contributions in the total query area in counts per square-kilometers. This endpoint does not support the deprecated ``types``, ``keys``, ``values`` parameters.
```

tyrasd (approved):

docs/endpoints.rst
FabiKo117: This wording would then not fit to the wording used in the other density description in line 134.

tyrasd: but this wording is the correct one :stuck_out_tongue: becuase a density of something divided by the area would have the unit of `thing / length^4` (instead of `thing / lenght^2`)

FabiKo117: Yeah I get now that you can also understand the old wording in a different way.. Then just fix also the old wording in the other line and it's fine :)

tyrasd (approved):

PR #139, opened 2021-02-22
round output coordinates to 7 decimal places

### LINKED ISSUE #138: Return real precision of coordinates
OSM uses a precision of 7 decimal values (https://wiki.openstreetmap.org/wiki/DE:Genauigkeit_von_Koordinaten). The ohsome API returns many coordinates with a precision of up to 14 decimals (enough to separate neighbouring atoms).

This is harmful in many ways

 - it suggest a higher precision to the user than the one actually available
 - it is considered bad practice ("indicates low quality processing"): https://gis.stackexchange.com/questions/8650/measuring-accuracy-of-latitude-and-longitude
 - can (and did) cause topological errors (ways sharing consecutive nodes no longer overlap but intersect) and invalid geometries (non-noded intersection): https://gis.stackexchange.com/questions/50399/fixing-non-noded-intersection-problem-using-postgis

FabiKo117:
### Description
limit the number of decimal values in the coordinates of response OSM features to 7

### Corresponding issue
fixes #138 

### Checklist

Please check all finished tasks. If some tasks do not apply to your PR, please cross their text out (by using `~...~`) and remove their checkboxes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ Changelog
 * update all tests using the filter parameter instead of deprecated types, keys, values ([#98])
 * update some default parameter values in swagger UI to slightly more sensible examples ([#113])
 * restructure packages and classes within the controller and output packages ([#117])
-* limit the length of coordinates in response OSM features to 7 ([#138])
+* round coordinates of returned OSM features to 7 decimal places ([#138])
 
 [#98]: https://github.com/GIScience/ohsome-api/issues/98
 [#111]: https://github.com/GIScience/ohsome-api/issues/111
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Changelog
 * update all tests using the filter parameter instead of deprecated types, keys, values ([#98])
 * update some default parameter values in swagger UI to slightly more sensible examples ([#113])
 * restructure packages and classes within the controller and output packages ([#117])
+* limit the length of coordinates in response OSM features to 7 ([#138])
 
 [#98]: https://github.com/GIScience/ohsome-api/issues/98
 [#111]: https://github.com/GIScience/ohsome-api/issues/111
@@ -25,6 +26,7 @@ Changelog
 [#117]: https://github.com/GIScience/ohsome-api/issues/117
 [#129]: https://github.com/GIScience/ohsome-api/issues/129
 [#131]: https://github.com/GIScience/ohsome-api/issues/131
+[#138]: https://github.com/GIScience/ohsome-api/issues/138
 
 
 ## 1.3.2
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ Changelog
 * update all tests using the filter parameter instead of deprecated types, keys, values ([#98])
 * update some default parameter values in swagger UI to slightly more sensible examples ([#113])
 * restructure packages and classes within the controller and output packages ([#117])
+* limit the length of coordinates in response OSM features to 7 ([#138])
```
tyrasd, line 20: hm, that's not quite correct. we are rounding to 7 decimal places, not _limiting the length to 7_. e.g. `49.4080159` has a "length" of 9 digits (or 10 characters when counting the `.`).

```suggestion
* round coordinates of returned OSM features to 7 decimal places ([#138])
```

src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/ExecutionUtils.java
```diff
@@ -958,6 +961,12 @@ private CSVWriter writeComments(HttpServletResponse servletResponse, List<String
     return properties;
   }
 
+  /** returns a new geometry out of the rounded coordinates of the given one */
+  private Geometry roundCoordsOfGeomToSevenDigits(Geometry oldGeom) {
+    PrecisionModel pm = new PrecisionModel(1E7);
+    return GeometryPrecisionReducer.reduce(oldGeom, pm);
```
tyrasd, line 967: a few comments:

* creating a new `PrecisionModel` for every time this method is called could introduce unwanted overhead here
* why not use a `GeometryPrecisionReducer` directly (`var gpr = new GeometryPrecisionReducer(… … return gpr.reduce(oldGeom);`), instead of using the static convenience method with a `PrecisionModel`?
* AFAICS, the `GeometryPrecisionReducer` will perform additional checks like eliminating duplicate vertices and checking polygons for validity:
  
  >  By default the reduced result is topologically valid (i.e. Geometry.isValid() is true). To ensure this a polygonal geometry is reduced in a topologically valid fashion (technically, by using snap-rounding). Note that this may change polygonal geometry structure (e.g. two polygons separated by a distance below the specified precision will be merged into a single polygon). 
  
  This could be problematic in two ways: it might introduce a performance penalty and it might alter the output in potentially unwanted ways (in general we want to return the OSM features' geometries AS closely to as they are mapped in OSM). Why not activate the [_pointwise_](https://locationtech.github.io/jts/javadoc/org/locationtech/jts/precision/GeometryPrecisionReducer.html#setPointwise-boolean-) and [_non-collapsing_](https://locationtech.github.io/jts/javadoc/org/locationtech/jts/precision/GeometryPrecisionReducer.html#setRemoveCollapsedComponents-boolean-) mode of the `GeometryPrecisionReducer`?

CHANGELOG.md
FabiKo117: I was thinking about "precision" but apparently wrote "length" for some reason 🙄 
I'll take your suggestion yeah.

tyrasd (approved):

PR #145, opened 2021-03-05
addition to fix #141

### LINKED ISSUE #141: Nullpointer Exception for /elements/geometry
The following message was visible in the logs upon an /elements/geometry request:

`Mar 01 10:55:02 heigitsv02 ignite.sh[9206]: Caused by: java.lang.NullPointerException`
`Mar 01 10:55:02 heigitsv02 ignite.sh[9206]: at org.heigit.ohsome.ohsomeapi.executor.ExecutionUtils.createOSMFeature(ExecutionUtils.java:387)`

using the following parameters:

```
bboxes: 7.109777450561525,46.00614082882865,7.187196731567385,46.04559279996195
filter: highway=track and (type:way)
time: 2007-11-01T00:00:00Z/2018-11-01T00:00:00Z/P1M
properties: tags,metadata
```

The issue is most likely caused by a given null as parameter instead of an empty EnumSet here:
https://github.com/GIScience/ohsome-api/blob/7a476ce54432bbe8cd7fc6429faa7dd487914251/src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/ElementsRequestExecutor.java#L128

FabiKo117:
### Description
replace null with empty enum set
addition to fix in pr #144 (issue #141)

### Checklist
~I have commented my code~
- [ ] I have added sufficient unit and API tests
~I have made corresponding changes to the [documentation](https://github.com/GIScience/ohsome-api/tree/master/docs)~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ Changelog
 * fix some invalid filters in the default swagger examples ([#111])
 * fix returning invalid GeoJSON using empty coordinates for deletion contributions ([#129], [#131])
 * fix using a proper boolean data type instead of a string for contributionType in response ([#135])
-* fix NPE with createOSMFeature ([#141])
+* fix null pointer exception with createOSMFeature ([#141], [#145])
 * make sure geometry filters are applied to all returned features of elementsFullHistory requests ([#109])
 
 ### Performance and Code Quality
@@ -37,6 +37,7 @@ Changelog
 [#135]: https://github.com/GIScience/ohsome-api/pull/135
 [#138]: https://github.com/GIScience/ohsome-api/issues/138
 [#141]: https://github.com/GIScience/ohsome-api/issues/141
+[#145]: https://github.com/GIScience/ohsome-api/pull/145
 
 
 ## 1.3.2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ Changelog
 * fix some invalid filters in the default swagger examples ([#111])
 * fix returning invalid GeoJSON using empty coordinates for deletion contributions ([#129], [#131])
 * fix using a proper boolean data type instead of a string for contributionType in response ([#135])
-* fix null pointer exception with createOSMFeature ([#141], [#145])
+* fix a null pointer exception in data extraction requests ([#141], [#145])
 * make sure geometry filters are applied to all returned features of elementsFullHistory requests ([#109])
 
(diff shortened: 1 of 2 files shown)


rtroilo (approved):

FabiKo117:
@rtroilo do we still want to maybe add a unit test regarding this issue? The thing is, if it should be an integration test, like most of the others, we'd have to find that problem within the Heidelberg.oshdb dataset, as that's the one used in our test environment.

joker234:
> @rtroilo do we still want to maybe add a unit test regarding this issue? The thing is, if it should be an integration test, like most of the others, we'd have to find that problem within the Heidelberg.oshdb dataset, as that's the one used in our test environment.

that would be very very cool, indeed :)

CHANGELOG.md
```diff
@@ -8,7 +8,7 @@ Changelog
 * fix some invalid filters in the default swagger examples ([#111])
 * fix returning invalid GeoJSON using empty coordinates for deletion contributions ([#129], [#131])
 * fix using a proper boolean data type instead of a string for contributionType in response ([#135])
-* fix NPE with createOSMFeature ([#141])
+* fix null pointer exception with createOSMFeature ([#141], [#145])
```
tyrasd, line 11: since `createOSMFeature` is an internal method name, I would suggest to use the following wording, for example:

```suggestion
* fix a null pointer exception in data extraction requests ([#141], [#145])
```

tyrasd:
> > @rtroilo do we still want to maybe add a unit test regarding this issue? The thing is, if it should be an integration test, like most of the others, we'd have to find that problem within the Heidelberg.oshdb dataset, as that's the one used in our test environment.
> 
> that would be very very cool, indeed :)

So, we're looking for elements which are unchanged in the time interval of the query which have an empty geometry, so that they trigger the NPE at `createOSMFeature(ExecutionUtils.java:387)`? The first condition might be easy to trigger since we can control the time interval of the test query, but not sure about the second one. Do you know which element(s) triggered the bug when you first encountered it?

tyrasd:
> So, we're looking for elements which are unchanged in the time interval of the query which have an empty geometry, so that they trigger the NPE at `createOSMFeature(ExecutionUtils.java:387)`?

I tried for a bit, but could not find any quick way to reproduce the bug with elements in the test data set. :unamused: maybe we can go ahead without a test for this specific bug?

tyrasd (dismissed):

tyrasd (approved):

PR #159, opened 2021-03-22
Template for issues

### LINKED ISSUE #153: Copy issue templates from other projects
Copying the issue templates from https://github.com/GIScience/oshdb to this project may improve the quality of reported issues.

bonaparten:
### Description
Add a template for issues and feature requests.

### Corresponding issue
Closes #153 

### Checklist
-

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -19,7 +19,7 @@ A clear and concise description of what the bug is.
 ### How to Reproduce
 Steps [e.g.] to reproduce the behaviour:
 1. Start locally the ohsome API
-2. Query this [URL] (http://localhost:8080/contributions/count?bboxes=8.67%2C49.39%2C8.71%2C49.42&time=2017-01-01%2F2017-01-02%2FP1Y)
+2. Send this request [URL] (http://localhost:8080/contributions/count?bboxes=8.67%2C49.39%2C8.71%2C49.42&time=2017-01-01%2F2017-01-02%2FP1Y)
 3. See error message 'Unable to find...'
 
 ### Expected Behaviour
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,37 @@
+---
+name: Bug report
+about: Create a report to help us improve the project.
+title: ''
+labels: 'bug'
+assignees: ''
+
+---
+
+### Bug Description
+A clear and concise description of what the bug is.
+
+### How to Reproduce
+Steps to reproduce the behaviour:
+1. Go to '...'
+2. Execute code '...'
+3. See error
+
+### Expected behaviour
+A clear and concise description of what you expected to happen.
+
+### Further Information
+#### Error messages, logs, screenshots
+If applicable, add printed error messages, log files or screenshots to help explain your problem.
+
+#### Additional Context
+Add any other context about the problem here.
+
+### System information
+Please complete the following information:
+ - OS: [e.g. Ubuntu 20.04 LTS]
+ - Java Version: [e.g. openjdk version "11.0.9.1"]
+ - ohsome API Version: [e.g. 1.4.0]
+ - Maven version: [e.g. 3.6.3]
+
+### Notes
+Please consider to upload a file if embedded request URLs, responses, error messages or logs are too long or bad displayed in your post.
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -31,5 +31,4 @@ If applicable, add printed error messages, log files or screenshots to help expl
 Add any other further information about the problem here.
 
 ### Notes
-1. Please consider to upload a file if request parameters, responses, error messages or logs are too long or poorly displayed in your bug report.
-2. Please consider to label your bug report [e.g. `priority:high`, `upstream`, etc.].
+Please consider to upload a file if request parameters, responses, error messages or logs are too long or poorly displayed in your bug report.
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -13,7 +13,7 @@ A clear and concise description of what the bug is.
 ### General Information
 Please include the following general information about the issue and list any additional steps needed to reproduce the bug.
  - Version of the ohsome API [e.g. `1.4.0`]
- - Which API instance was requested [e.g. remote instance, local instance with an h2 file or to the cluster connected, ohsome dashboard, etc.]
+ - Which API instance was requested [e.g. remote instance at `https://api.ohsome.org/v1`, local instance with an h2 file, etc.]
  - Affected endpoint(s) [e.g. /contributions/count]
  - URL of your request (and request body if applicable)
  - Used HTTP method [GET or POST]
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -32,4 +32,4 @@ Add any other further information about the problem here.
 
 ### Notes
 1. Please consider to upload a file if request parameters, responses, error messages or logs are too long or poorly displayed in your bug report.
-2. Please consider to label your bug report [e.g. priority:high, upstream, etc.].
+2. Please consider to label your bug report [e.g. `priority:high`, `upstream`, etc.].
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -31,4 +31,5 @@ If applicable, add printed error messages, log files or screenshots to help expl
 Add any other further information about the problem here.
 
 ### Notes
-Please consider to upload a file if request parameters, responses, error messages or logs are too long or poorly displayed in your bug report.
+1. Please consider to upload a file if request parameters, responses, error messages or logs are too long or poorly displayed in your bug report.
+2. Please consider to label your bug report [e.g. priority:high, upstream, etc.].
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -11,15 +11,15 @@ assignees: ''
 A clear and concise description of what the bug is.
 
 ### General Information
-Affected endpoint(s) [e.g. /contributions/count].
-Affected parameter(s) [e.g. bpolys, format].
-HTTP method used [e.g. POST].
-Used tool/library for the request [e.g. cURL, Postman].
+ - Affected endpoint(s) [e.g. /contributions/count]
+ - Affected parameter(s) [e.g. bpolys, format]
+ - HTTP method used [e.g. POST]
+ - Used tool/library for the request [e.g. cURL, Postman]
 
 ### How to Reproduce
 Steps [e.g.] to reproduce the behaviour:
 1. Start locally the ohsome API
-2. Query the URL 'http://localhost:8080/contributions/count?bboxes=8.67%2C49.39%2C8.71%2C49.42&time=2017-01-01%2F2017-01-02%2FP1Y'
+2. Query this [URL] (http://localhost:8080/contributions/count?bboxes=8.67%2C49.39%2C8.71%2C49.42&time=2017-01-01%2F2017-01-02%2FP1Y)
 3. See error message 'Unable to find...'
 
 ### Expected Behaviour
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -13,7 +13,7 @@ A clear and concise description of what the bug is.
 ### General Information
 Please include the following general information about the issue and list any additional steps needed to reproduce the bug.
  - Version of the ohsome API [e.g. `1.4.0`]
- - Which API instance was requested [e.g. remote instance at `https://api.ohsome.org/v1`, local instance with an h2 file, etc.]
+ - Which API instance was requested [e.g. remote instance at `https://api.ohsome.org/v1`, local instance with an h2 file or connected to the cluster, etc.]
  - Affected endpoint(s) [e.g. /contributions/count]
  - URL of your request (and request body if applicable)
  - Used HTTP method [GET or POST]
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -13,9 +13,9 @@ A clear and concise description of what the bug is.
 ### General Information
 Please include the following general information about the issue and list any additional steps needed to reproduce the bug.
  - Version of the ohsome API [e.g. `1.4.0`]
- - Which API instance was requested [e.g. `https://api.ohsome.org/v1`]
+ - Which API instance was requested [e.g. remote instance, local instance with an h2 file or to the cluster connected, ohsome dashboard, etc.]
  - Affected endpoint(s) [e.g. /contributions/count]
- - URL of your request
+ - URL of your request (and request body if applicable)
  - Used HTTP method [GET or POST]
  - Utilized tool/library for the request [e.g. cURL, Postman, ohsome-py, etc.]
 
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,37 @@
...
+### How to Reproduce
+Steps to reproduce the behaviour:
+1. Go to '...'
+2. Execute code '...'
+3. See error
+
+### Expected behaviour
+A clear and concise description of what you expected to happen.
+
+### Further Information
```
joker234, line 22: I would really like to alter this section in contrast to the [oshdb](https://github.com/GIScience/oshdb/blob/master/.github/ISSUE_TEMPLATE/bug_report.md) a little bit. It would be nice to have some kind of template with _endpoint_, _parameters_, _data_(?), _used requesting tool/library_ … in it to remind people that it's easier to bugfix if we have the important information.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,37 @@
...
+labels: 'bug'
+assignees: ''
+
+---
+
+### Bug Description
+A clear and concise description of what the bug is.
+
+### How to Reproduce
+Steps to reproduce the behaviour:
```
joker234, line 14: This should be an example. Most people might experience issues when requesting something.

joker234 (changes requested):
apart from the two comments good, thanks :+1:

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -10,23 +10,30 @@ assignees: ''
 ### Bug Description
 A clear and concise description of what the bug is.
 
+### General Information
+Affected endpoint(s) [e.g. /contributions/count].
+Affected parameter(s) [e.g. bpolys, format].
+HTTP method used [e.g. POST].
+Used tool/library for the request [e.g. cURL, Postman].
```
joker234, line 17: Please have a look at the [rendered version](https://github.com/GIScience/ohsome-api/blob/issues-template/.github/ISSUE_TEMPLATE/bug_report.md#general-information) of this markdown file as it gives an indicator how it looks like in the later issue. I would suggest to use bulletpoints as within _System Information_.

joker234 (dismissed):

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,44 @@
...
+assignees: ''
+
+---
+
+### Bug Description
+A clear and concise description of what the bug is.
+
+### General Information
+ - Affected endpoint(s) [e.g. /contributions/count]
+ - Affected parameter(s) [e.g. bpolys, format]
```
tyrasd, line 15:
```suggestion
 - Used query parameter(s) [bpolys/bboxes/bcircles, filter, time, format, etc.]
```

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,44 @@
...
+ - Used tool/library for the request [e.g. cURL, Postman]
+
+### How to Reproduce
+Steps [e.g.] to reproduce the behaviour:
+1. Start locally the ohsome API
+2. Send this request [URL] (http://localhost:8080/contributions/count?bboxes=8.67%2C49.39%2C8.71%2C49.42&time=2017-01-01%2F2017-01-02%2FP1Y)
+3. See error message 'Unable to find...'
+
+### Expected Behaviour
+A clear and concise description of what you expected to happen [e.g. "The error message should be..."]
```
tyrasd, line 26:
```suggestion
A clear and concise description of what you expected to happen.
```

I would drop the example here, since the expected behaviour could be so many different things. Similarly we also don't give an example for the initial _bug description_ section.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,44 @@
...
+
+#### Additional information
+Add any other further information about the problem here.
+
+### System Information
+Please complete the following information:
+ - OS: [e.g. Ubuntu 20.04 LTS]
+ - Java Version: [e.g. openjdk version "11.0.9.1"]
+ - ohsome API Version: [e.g. 1.4.0]
+ - Maven version: [e.g. 3.6.3]
```
tyrasd, line 41:
```suggestion
```

This is generally irrelevant for ohsome API issues, so I would not ask for it in every bug report. If needed, the users can add such information in the _additional information_ section.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,44 @@
...
+
+### Expected Behaviour
+A clear and concise description of what you expected to happen [e.g. "The error message should be..."]
+
+### Further Information
+
+#### Error messages, logs, screenshots
+If applicable, add printed error messages, log files or screenshots to help explain your problem.
+
+#### Additional information
```
tyrasd, line 33:
```suggestion
#### Additional Information
```

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,44 @@
...
+1. Start locally the ohsome API
+2. Send this request [URL] (http://localhost:8080/contributions/count?bboxes=8.67%2C49.39%2C8.71%2C49.42&time=2017-01-01%2F2017-01-02%2FP1Y)
+3. See error message 'Unable to find...'
+
+### Expected Behaviour
+A clear and concise description of what you expected to happen [e.g. "The error message should be..."]
+
+### Further Information
+
+#### Error messages, logs, screenshots
```
tyrasd, line 30:
```suggestion
#### Error Messages, Logs, Screenshots
```

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,44 @@
...
+
+### System Information
+Please complete the following information:
+ - OS: [e.g. Ubuntu 20.04 LTS]
+ - Java Version: [e.g. openjdk version "11.0.9.1"]
+ - ohsome API Version: [e.g. 1.4.0]
+ - Maven version: [e.g. 3.6.3]
+
+### Notes
+Please consider to upload a file if embedded request URLs, responses, error messages or logs are too long or bad displayed in your post.
```
tyrasd, line 44:
```suggestion
> Please consider to upload a file if request parameters, responses, error messages or logs are too long or poorly displayed in your bug report.
```

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,20 @@
...
+name: Feature request
+about: Suggest an idea or enhancement to the ohsome API.
+title: ''
+labels: 'feature request'
+assignees: ''
+
+---
+
+### Problem Description
+Please add a clear and concise description of what the problem is, e.g. "I'm always frustrated when..., because..."
```
tyrasd, line 11: I don't like the negative connotation in this section too much. Most of the time feature requests are not about problems (that's what a bug report is for), but instead focus on potential improvements. 

For feature request, users often forget to state their intention/rationale behind suggested ideas. This makes it often hard to evaluate feature requests objectively. I would prompt the user to try to formulate their suggestion as a user story, perhaps like this:

```suggestion
### Feature Description
Please add a clear and concise description of what your idea is, what use cases it solves. You could try to formulate it as a user story using the following template: _“As a … I would like to … so that …”_ (e.g., _“As a researcher I would like to request the daily number of OSM changesets touching particular map features in a given area in order to analyse trends in mapping activity.”_).
```

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,20 @@
...
+
+### Expected Solution
+Describe the solution you'd like
+
+### Alternative Solutions
+Describe alternatives you've considered
```
tyrasd, line 17:
```suggestion
### Expected Solution
Describe the solution you would like to see.

### Alternative Solutions
Describe alternatives you have considered.
```

tyrasd (commented):
See some suggestions inline.

.github/ISSUE_TEMPLATE/bug_report.md
bonaparten: Since it is a subparagraph I would use capitalization only for the first word.

.github/ISSUE_TEMPLATE/bug_report.md
bonaparten: Same here

.github/ISSUE_TEMPLATE/feature_request.md
bonaparten: I don't like to put user story models here. I think the mentioned example protracts too much the description and it isn't very useful for the users, since most of the features requested here are small and specific enhancements.

.github/ISSUE_TEMPLATE/bug_report.md
tyrasd: Why the distinction between "normal" headings and headings of subsections? I would recommend to always use one single consistent style for headings throughout a text.

.github/ISSUE_TEMPLATE/feature_request.md
tyrasd: ok, user-story style might be overkill, fair enough. But my point was that the current wording isn't about _enhancements_ at all. It reads more like a bug report, talking about _problems_ and _frustration_.

Can you come up with a better wording which is tailored towards _small and specific enhancements_?

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,20 @@
+---
+name: Feature request
+about: Suggest an enhancement to the ohsome API.
+title: ''
+labels: 'feature request'
```
tyrasd, line 5: currently, we don't have such a [label](https://github.com/GIScience/ohsome-api/labels). I think `enhancement` would be used?!

```suggestion
labels: 'enhancement'
```

tyrasd (changes requested):
AFAIK, only collaborators can edit the labels of an issue (except for the pre-assigned basic "issue type" label from the respective template, e.g., `bug`). Therefore, I would suggest to drop the sections requesting the addition of labels.

.github/ISSUE_TEMPLATE/feature_request.md
bonaparten: `enhancement` is also ok since the description of the `enhancement` label says `New feature or request`

bonaparten: > AFAIK, only collaborators can edit the labels of an issue (except for the pre-assigned basic "issue type" label from the respective template, e.g., `bug`). Therefore, I would suggest to drop the sections requesting the addition of labels.

Having the possibility to set the priority was a request by @SlowMo24, since we use labels I thought just to draw attention to the use of the labels. No problem about deleting this part. @SlowMo24 what do you think?

bonaparten: Same for the bug report

SlowMo24: I'm fine with anything but having some sort of 'how urgent' information given by the user **may** be of value. Just to see the difference between 'I need this for my research tomorrow' and 'That would also be great, but I wont use it til next year and there is a workaround'. It should though be clear that we have our own prioritisation and obey noone ;-)

SlowMo24 (commented):
> Same for the bug report

I'm not sure because a bug should always be fixed so the priority starts at 'high':'extreme' while for feature requests it goes from 'won't implement':'high'

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,20 @@
...
+---
+
+### Use Case Description
+Describe the use case which would benefit from this enhancement.
+
+### Request Description
+Add a clear description about the enhancement. Please be specific about the features you would like to have.
+
+### Additional Information
+Add any other information, examples, screenshots, etc. about the feature request here.
```
tyrasd, line 17: to address https://github.com/GIScience/ohsome-api/pull/159#discussion_r608502509, let's perhaps write it like this:

```suggestion
### Request Description
Add a clear description about the enhancement. Please be specific about the feature you would like to have.

### Additional Information
Add any other information, examples, screenshots, possible workarounds, etc. about the feature request here. You may indicate here how important the enhacement is for your use case and whether or not you would be willing to support the development (from coding a pull request, to helping as a beta tester).
```

?

bonaparten:
> > Same for the bug report
> 
> I'm not sure because a bug should always be fixed so the priority starts at 'high':'extreme' while for feature requests it goes from 'won't implement':'high'



> > Same for the bug report
> 
> I'm not sure because a bug should always be fixed so the priority starts at 'high':'extreme' while for feature requests it goes from 'won't implement':'high'

I meant that we have the same problem with the labeling by non-collaborators also for the bug template.

.github/ISSUE_TEMPLATE/feature_request.md
bonaparten: That is fine, but I think the last part "and whether or not..." it doesn't add any important information since we code by ourselves. If we need for some help (e.g. a beta tester) we can just ask him/her.

tyrasd: Well, it's an open source project after all. So it could be the case that at some point someone external might want to contribute. :)
Or, for example, if let's say that @SlowMo24 really needs a certain feature very urgently ( :wink: ), it could speed up some thinks if we know that we can count on active support.

Or if a feature request comes from a team member, but who has not time/skill/… to implement a certain feature, it would also be good to know, IMO.

bonaparten: I see the point, just I don't think the text adds really value. But I have no problem with adding this text.;-)

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,34 @@
...
+
+### Bug Description
+A clear and concise description of what the bug is.
+
+### General Information
+Please include the following general information about the issue and list any additional steps needed to reproduce the bug.
+ - Version of the ohsome API [e.g. `1.4.0`]
+ - Which API instance was requested [e.g. `https://api.ohsome.org/v1`]
+ - Affected endpoint(s) [e.g. /contributions/count]
+ - URL of your request
```
tyrasd, line 18: one last point: since we already asked for the api instance and endpoint, there's no need to ask for the URL here again (at least for POST requests). What we however need to know are the used parameters:

```suggestion
 - parameters used for your request
```

tyrasd (changes requested):
one last point from my side :blush:

.github/ISSUE_TEMPLATE/bug_report.md
bonaparten: I think it is useful to know the URL, at least it would be fast for us to just copy/paste and check the bug (if any). Errors could also come because of a bad request and not because of a bug, e.g. filter syntax used in the wrong way.
Instead of that, I would just change the text and get rid of URLs in the point about the API instance:
 - Which API instance was requested [e.g. remote instance, locale instance with an h2 file or to the cluster connected, ohsome dashboard, etc.]
 - URL of your request (and request body if applicable)

tyrasd: sounds good

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -13,9 +13,9 @@ A clear and concise description of what the bug is.
 ### General Information
 Please include the following general information about the issue and list any additional steps needed to reproduce the bug.
  - Version of the ohsome API [e.g. `1.4.0`]
- - Which API instance was requested [e.g. `https://api.ohsome.org/v1`]
+ - Which API instance was requested [e.g. remote instance, local instance with an h2 file or to the cluster connected, ohsome dashboard, etc.]
```
tyrasd, line 16: IMHO it's important to know _which_ remote instance was used, so I would not skip that. I'd drop the local+cluster and dashboard examples because two examples are enough and local+cluster is a quite rare/specific set-up and the dashboard is just using to the default remote instance.

```suggestion
 - Which API instance was requested [e.g. remote instance at `https://api.ohsome.org/v1`, local instance with an h2 file, etc.]
```

Or, if you prefer to keep the third example, it would sound better if we write: “_… or connected to the cluster, …_”

FabiKo117: The URL should definitely be included there as an example, as it's most likely the main access point. The direct cluster connection is indeed quite specific, but still worth mentioning I guess.

tyrasd (approved):

PR #161, opened 2021-03-22
avoid unneccessary geometry building during count/ratio requests

tyrasd:
### Description

Improve performance of `count` queries by skipping calling `getGeometry` in the `count` code paths. This eliminates the (sometimes expensive) construction of JTS geometries in cases where it is not actually needed.

### Todo

* [x] perform some simple benchmarks to see how big of a difference this makes

### Checklist

### DIFF
--- a/src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/AggregateRequestExecutor.java
+++ b/src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/AggregateRequestExecutor.java
@@ -81,31 +81,27 @@ public class AggregateRequestExecutor extends RequestExecutor {
     final SortedMap<OSHDBTimestamp, ? extends Number> result;
     MapReducer<OSMEntitySnapshot> mapRed = null;
     mapRed = inputProcessor.processParameters();
+    var mapRedGeom = mapRed.map(OSMEntitySnapshot::getGeometry);
     switch (requestResource) {
       case COUNT:
         result = mapRed.aggregateByTimestamp().count();
         break;
-      case AREA:
-        result = mapRed.aggregateByTimestamp()
-            .sum((SerializableFunction<OSMEntitySnapshot, Number>) snapshot -> ExecutionUtils
-                .cacheInUserData(snapshot.getGeometry(), () -> Geo.areaOf(snapshot.getGeometry())));
-        break;
-      case LENGTH:
-        result = mapRed.aggregateByTimestamp()
-            .sum((SerializableFunction<OSMEntitySnapshot, Number>) snapshot -> ExecutionUtils
-                .cacheInUserData(snapshot.getGeometry(),
-                    () -> Geo.lengthOf(snapshot.getGeometry())));
-        break;
       case PERIMETER:
-        result = mapRed.aggregateByTimestamp()
-            .sum((SerializableFunction<OSMEntitySnapshot, Number>) snapshot -> {
-              if (snapshot.getGeometry() instanceof Polygonal) {
-                return ExecutionUtils.cacheInUserData(snapshot.getGeometry(),
-                    () -> Geo.lengthOf(snapshot.getGeometry().getBoundary()));
-              } else {
+        result = mapRedGeom.aggregateByTimestamp()
+            .sum(geom -> {
+              if (!(geom instanceof Polygonal)) {
                 return 0.0;
               }
+              return ExecutionUtils.cacheInUserData(geom, () -> Geo.lengthOf(geom.getBoundary()));
             });
+        break;
+      case LENGTH:
+        result = mapRedGeom.aggregateByTimestamp()
+            .sum(geom -> ExecutionUtils.cacheInUserData(geom, () -> Geo.lengthOf(geom)));
+        break;
+      case AREA:
+        result = mapRedGeom.aggregateByTimestamp()
+            .sum(geom -> ExecutionUtils.cacheInUserData(geom, () -> Geo.areaOf(geom)));
         break;
       default:
         throw new RuntimeException("Unsupported RequestResource type for this processing. "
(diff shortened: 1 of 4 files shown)


src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/ElementsRequestExecutor.java
```diff
@@ -684,8 +684,8 @@ public static Response aggregateRatio(RequestResource requestResource,
     mapRed = mapRed.filter(combinedFilter);
     var preResult = mapRed.aggregateByTimestamp().aggregateBy(snapshot -> {
       OSMEntity entity = snapshot.getEntity();
-      boolean matches1 = filterExpr1.applyOSMGeometry(entity, snapshot.getGeometry());
-      boolean matches2 = filterExpr2.applyOSMGeometry(entity, snapshot.getGeometry());
+      boolean matches1 = filterExpr1.applyOSMGeometry(entity, snapshot::getGeometry);
```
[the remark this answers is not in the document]
tyrasd: `.` executes the method immediately (giving a `Geometry` object to `applyOSMGeometry`), while `::` just hands a "pointer" to the `getGeometry` method to `applyOSMGeometry`, which then only executes it when it is really necessary to do so (e.g. when the filter includes a geometry type filter). So, when the filter of a `count` query does not need to build the geometry, this change would then skip the not-neccessary call of `getGeometry`.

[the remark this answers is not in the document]
FabiKo117: Ahh I see, makes sense then. Could you already check if this also results in a performance increase?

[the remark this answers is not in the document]
tyrasd: not yet, but it definitely makes sense to do some basic benchmarking for this :+1:

src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/AggregateRequestExecutor.java
```diff
@@ -81,32 +81,28 @@ public Response aggregate() throws Exception {
...
+        result = mapRedGeom.aggregateByTimestamp()
+            .sum(geom -> ExecutionUtils.cacheInUserData(geom, () -> Geo.lengthOf(geom)));
+        break;
+      case AREA:
+        result = mapRedGeom.aggregateByTimestamp()
+            .sum(geom -> ExecutionUtils.cacheInUserData(geom, () -> Geo.areaOf(geom)));
+        break;
       default:
         throw new RuntimeException("Unsupported RequestResource type for this processing. "
             + "Only COUNT, LENGTH, PERIMETER, and AREA are permitted here");
```
joker234, line 108: Is there a reason you re-sorted the cases?

src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/ExecutionUtils.java
```diff
@@ -443,28 +442,26 @@ public void writeCsvResponse(Result[] resultSet, HttpServletResponse servletResp
   public static <K extends Comparable<K> & Serializable, V extends Number>
       SortedMap<OSHDBCombinedIndex<OSHDBTimestamp, K>, V> computeResult(
       RequestResource requestResource,
-      MapAggregator<OSHDBCombinedIndex<OSHDBTimestamp, K>, OSMEntitySnapshot> preResult)
+      MapAggregator<OSHDBCombinedIndex<OSHDBTimestamp, K>, OSMEntitySnapshot> mapAgg)
```
joker234, line 445: I'm not 100% sure, but shouldn't these two lines be indented one level more, as they are parameters to an additional function?

src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/ExecutionUtils.java
```diff
@@ -443,28 +442,26 @@ public void writeCsvResponse(Result[] resultSet, HttpServletResponse servletResp
...
+            .sum(geom -> cacheInUserData(geom, () -> Geo.lengthOf(geom)));
       case AREA:
-        return (SortedMap<OSHDBCombinedIndex<OSHDBTimestamp, K>, V>) preResult
-            .sum((SerializableFunction<OSMEntitySnapshot, Number>) snapshot -> cacheInUserData(
-                snapshot.getGeometry(), () -> Geo.areaOf(snapshot.getGeometry())));
+        return (SortedMap<OSHDBCombinedIndex<OSHDBTimestamp, K>, V>) mapAggGeom
+            .sum(geom -> cacheInUserData(geom, () -> Geo.areaOf(geom)));
       default:
         return null;
     }
```
joker234, line 467: same question about resorting as before

joker234 (commented):
looks pretty good :+1: Just some minor remarks/questions.

src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/AggregateRequestExecutor.java
tyrasd: if I remember correctly, I did this to make the order the same for all of these very similar `switch (aggregationType)` statements. But apparently, I overlooked one which still has `LENGTH` and `PERIMETER` reversed.

src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/ExecutionUtils.java
tyrasd: At first I'd thought your are right. However, checkstyle doesn't complain about it, so maybe both options are valid? Also, since some other long functions are formatted this way in the ohsome-api I'd rather not touch them in this PR. :shrug:

joker234 (approved):

PR #162, opened 2021-03-23
Add 'contributionTypes' property to 'properties' parameter

### LINKED ISSUE #136: Add property 'contributionTypes' to 'properties' parameter
Currently, the `@contributionType` field gets added to the response features of the `/contributions/{geometryType}` endpoints if the `properties=metadata` parameter is given in the request. However, as it is not directly related to the other fields that get added via setting this parameter, it would be better to have it as a distinct value of the `properties` parameter, so that it gets added via setting `properties=contributionTypes` (or alternatively `contribution types`). This property should only be allowed for the `/contributions/{geometryType}` endpoint.

bonaparten:
### Description
Add the 'contributionTypes' property to property parameter for `contribution{geometryType} `endpoints.

### Corresponding issue
Closes #136 

### Checklist

### DIFF
--- a/src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/ExecutionUtils.java
+++ b/src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/ExecutionUtils.java
@@ -382,8 +382,8 @@ public class ExecutionUtils {
   /** Creates the <code>Feature</code> objects in the OSM data response. */
   public org.wololo.geojson.Feature createOSMFeature(OSMEntity entity, Geometry geometry,
       Map<String, Object> properties, Set<Integer> keysInt, boolean includeTags,
-      boolean includeOSMMetadata, boolean isContributionsEndpoint, ElementsGeometry elemGeom,
-      EnumSet<ContributionType> contributionTypes) {
+      boolean includeOSMMetadata, boolean includeContributionTypes, boolean isContributionsEndpoint,
+      ElementsGeometry elemGeom, EnumSet<ContributionType> contributionTypes) {
     if (geometry.isEmpty() && !contributionTypes.contains(ContributionType.DELETION)) {
       // skip invalid geometries (e.g. ways with 0 nodes)
       return null;
@@ -395,8 +395,15 @@ public class ExecutionUtils {
           .filter(t -> keysInt.contains(t.getKey())).toArray(OSHDBTag[]::new));
     }
     if (includeOSMMetadata) {
-      properties =
-          addAdditionalProperties(entity, properties, isContributionsEndpoint, contributionTypes);
+      properties.put("@version", entity.getVersion());
+      properties.put("@osmType", entity.getType());
+      properties.put("@changesetId", entity.getChangesetId());
+      if (isContributionsEndpoint) {
+      properties = addContributionTypes(properties, contributionTypes);
+    }
+    }
+    if (includeContributionTypes && !includeOSMMetadata) {
+      properties = addContributionTypes(properties, contributionTypes);
     }
     properties.put("@osmId", entity.getType().toString().toLowerCase() + "/" + entity.getId());
     GeoJSONWriter gjw = new GeoJSONWriter();
@@ -938,14 +945,9 @@ public class ExecutionUtils {
     return writer;
   }
 
-  /** Adds additional properties like the version or the changeset ID to the feature. */
-  private Map<String, Object> addAdditionalProperties(OSMEntity entity,
-      Map<String, Object> properties, boolean isContributionsEndpoint,
+  /** Adds contribution types properties like creation to the feature. */
+  private Map<String, Object> addContributionTypes(Map<String, Object> properties,
       EnumSet<ContributionType> contributionTypes) {
-    properties.put("@version", entity.getVersion());
-    properties.put("@osmType", entity.getType());
-    properties.put("@changesetId", entity.getChangesetId());
-    if (isContributionsEndpoint) {
       if (contributionTypes.contains(ContributionType.CREATION)) {
         properties.put("@creation", true);
       }
@@ -958,7 +960,6 @@ public class ExecutionUtils {
       if (contributionTypes.contains(ContributionType.GEOMETRY_CHANGE)) {
         properties.put("@geometryChange", true);
       }
-    }
     return properties;
   }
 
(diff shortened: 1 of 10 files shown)


src/main/lombok/org/heigit/ohsome/ohsomeapi/inputprocessing/InputProcessor.java
```diff
@@ -391,28 +392,44 @@ public boolean compareKeysValues(String[] keys, String[] keys2, String[] values,
...
+        } else if (oldUnclippedParameter) {
+          this.clipGeometry = false;
+        } else if ("contributionTypes".equalsIgnoreCase(property)) {
+          this.includeContributionTypes = true;
+        } else {
+          throw new BadRequestException(ExceptionMessages.PROPERTIES_PARAM_CONTR);
+        }
       }
     }
   }
```
FabiKo117, line 435: I think you can write this method in an easier way with just one for-each loop. Please take a look at my suggested approach and check, but I think it should also work like that. At least the local tests worked fine.

```java
public void processPropertiesParam() {
    String[] properties =
        splitParamOnComma(createEmptyArrayIfNull(requestParameters.get("properties")));
    for (String property : properties) {
      @Deprecated
      boolean oldUnclippedParameter = "unclipped".equalsIgnoreCase(property);
      if ("tags".equalsIgnoreCase(property)) {
        this.includeTags = true;
      } else if ("metadata".equalsIgnoreCase(property)) {
        this.includeOSMMetadata = true;
      } else if (oldUnclippedParameter) {
        this.clipGeometry = false;
      } else if (RequestUtils.isContributionsExtraction(requestUrl)) {
        if ("contributionTypes".equalsIgnoreCase(property)) {
          this.includeContributionTypes = true;
        } else {
          throw new BadRequestException(ExceptionMessages.PROPERTIES_PARAM_CONTR);
        }
      } else {
        throw new BadRequestException(ExceptionMessages.PROPERTIES_PARAM);
      }
    }
  }
```

src/test/java/org/heigit/ohsome/ohsomeapi/controller/DataExtractionTest.java
```diff
@@ -129,7 +130,18 @@ public void elementsBboxTest() {
     assertEquals("Polygon", featureGeom.get("type").asText());
     assertEquals(5, featureGeom.get("coordinates").get(0).size());
   }
-
+  
+  @Test
+  public void contributionTypesValueElementsBboxTest() {
```
FabiKo117, line 135: I'd prefer to use a more meaningful naming scheme for the tests. I've started it with the tests for `/contributions/count` as you can see for example with this test:
https://github.com/GIScience/ohsome-api/blob/a177dc3d3ccb6c74ccfaa9a117925271e3ffee76/src/test/java/org/heigit/ohsome/ohsomeapi/controller/PostControllerTest.java#L1144
When a general re-organization of the tests would be performed at some point, it would be great to apply such a scheme to all test methods. Am also open to discuss about some other standards regarding the naming of test methods.

FabiKo117 (changes requested):
See the in-line comments and suggestions.

src/test/java/org/heigit/ohsome/ohsomeapi/controller/DataExtractionTest.java
```diff
@@ -446,7 +458,19 @@ public void contributionsGeometryCoordinateTest() {
         + "&filter=id:node/3429511451&time=2017-01-01,2019-01-01", JsonNode.class);
     JsonNode feature = Helper.getFeatureByIdentifier(response, "@osmId", "node/3429511451");
     assertEquals(49.418466, feature.get("geometry").get("coordinates").get(1).asDouble(), 0);
-  }
+    
+  }
+    @Test
+    public void contributionTypesTest() {
```
FabiKo117, line 464: Only one thing I have to point out: as it's very likely that we will have a `/groupBy/contributionType` endpoint at some point, it would be better to make it clear in the name of this test that it's about the parameter `properties=contributionTypes`.

```suggestion
    public void contributionTypesPropertiesParameterTest() {
```

FabiKo117 (changes requested):
Only one inline comment. Rest looks good 👍

src/test/java/org/heigit/ohsome/ohsomeapi/controller/DataExtractionTest.java
bonaparten: Yes, you are right!

FabiKo117 (approved):

src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/ExecutionUtils.java
```diff
@@ -395,8 +395,15 @@ public void writeCsvResponse(Result[] resultSet, HttpServletResponse servletResp
...
     }
     if (includeOSMMetadata) {
-      properties =
-          addAdditionalProperties(entity, properties, isContributionsEndpoint, contributionTypes);
+      properties.put("@version", entity.getVersion());
+      properties.put("@osmType", entity.getType());
+      properties.put("@changesetId", entity.getChangesetId());
+      if (isContributionsEndpoint) {
+      properties = addContributionTypes(properties, contributionTypes);
+    }
```
tyrasd, line 403: aahm… what happened to the indentation here? :see_no_evil: PS: please don't just tick off the _My code follows the code-style rules_ checkbox without actually checking it :stuck_out_tongue:

tyrasd: oops, there were quite a few of them all over the place here. See https://jenkins.ohsome.org/job/ohsome-api/job/master/508/checkstyle/new/. I've opened https://github.com/GIScience/ohsome-api/pull/173 to hopefully fix all of them.

PS: please install and use the checkstyle plugin in your [IDE](https://github.com/GIScience/ohsome-api/blob/master/CONTRIBUTING.md#code-style) and make it a habit to run it befor submitting a PR. :pray: 

PSS: please consider configuring your IDE to remove superfluous trailing whitespace at the end of lines, like [below](https://github.com/GIScience/ohsome-api/pull/162/files#diff-9175ccbae42fea93a094d76cb63b8f7cc0f5a58ea80a48615cd5d57ce6f3dc6eR142) ([fix](https://github.com/GIScience/ohsome-api/pull/173/files#diff-9175ccbae42fea93a094d76cb63b8f7cc0f5a58ea80a48615cd5d57ce6f3dc6eL142-R142)).

PR #168, opened 2021-03-25
swagger tweaks: reordering and remove deprecated parameters

tyrasd:
### Description
A few changes to the swagger api playground:

* reorder "tags" in specs:
  * for aggregation resources: `count` first, then `length`, `area` and `perimeter`, then `users` and last `contributions`
  * for data extraction: `geometry`, then `elementsFullHistory/geometry`, then `contributions/geometry`
  * no change for metadata :wink: 
  * reasoning: alphabetical sorting is not the most user friendly, IMHO. I propose the following sorting: simpler, more often used resources should come first, and more complex ones later
* start phasing out deprecated types/keys/values parameters by not offering them in swagger anymore
  * reasoning: a) it makes it more evident that these parameters should not be used, b) we already have a few endpoint which do _not_ support these old deprecated filters (e.g. all `contributions` endpoints), yet swagger still shows them as apparently valid options.
* ~~make time parameter for contribution endpoint optional (fall back to full time range if not specified??) TODO: check if this is a good idea and if it even works as intended~~

### Checklist

### DIFF
--- a/src/main/lombok/org/heigit/ohsome/ohsomeapi/config/SwaggerConfig.java
+++ b/src/main/lombok/org/heigit/ohsome/ohsomeapi/config/SwaggerConfig.java
@@ -177,15 +177,6 @@ public class SwaggerConfig implements SwaggerResourcesProvider {
     globalOperationParams.add(new ParameterBuilder().name("bpolys")
         .description(ParameterDescriptions.BPOLYS).modelRef(new ModelRef(string))
         .parameterType(query).defaultValue("").required(false).build());
-    globalOperationParams.add(new ParameterBuilder().name("types")
-        .description(ParameterDescriptions.DEPRECATED_USE_FILTER).modelRef(new ModelRef(string))
-        .allowMultiple(true).parameterType(query).defaultValue("").required(false).build());
-    globalOperationParams.add(new ParameterBuilder().name("keys")
-        .description(ParameterDescriptions.DEPRECATED_USE_FILTER).modelRef(new ModelRef(string))
-        .parameterType(query).defaultValue("").required(false).build());
-    globalOperationParams.add(new ParameterBuilder().name("values")
-        .description(ParameterDescriptions.DEPRECATED_USE_FILTER).modelRef(new ModelRef(string))
-        .parameterType(query).defaultValue("").required(false).build());
     globalOperationParams
         .add(new ParameterBuilder().name("filter").description(ParameterDescriptions.FILTER)
             .modelRef(new ModelRef(string)).parameterType(query)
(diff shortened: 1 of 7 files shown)


FabiKo117:
I agree with both points: bringing a specific ordering in the speccs and further removing the deprecated `types, keys, values` parameters. This also reduces the code further. 

What just came into my mind: Would it make sense to include a specific message (e.g. in a dedicated JSON field) in the response if the user used a deprecated parameter, endpoint, whatever to get this response and that he/she should check the docs and update this, as this specific request won't work anymore after the next major release? To better display this, we could add a specific section in the docs explaining which parameters, endpoints, etc. were deprecated in which version and from when they won't be available anymore.

tyrasd:
> Would it make sense to include a specific message (e.g. in a dedicated JSON field) in the response if the user used a deprecated parameter, endpoint, whatever to get this response and that he/she should check the docs and update this, as this specific request won't work anymore after the next major release?

That sounds like a good idea! :+1: At least for the JSON (aggregation) output format this should be possible without a problem. I'm not 100% sure about csv (maybe including it in the header comment would work?) and GeoJSON, though.

src/main/lombok/org/heigit/ohsome/ohsomeapi/config/SwaggerConfig.java
```diff
@@ -174,15 +177,6 @@ private ApiInfo apiInfo() {
...
         .parameterType(query).defaultValue("").required(false).build());
-    globalOperationParams.add(new ParameterBuilder().name("types")
-        .description(ParameterDescriptions.DEPRECATED_USE_FILTER).modelRef(new ModelRef(string))
-        .allowMultiple(true).parameterType(query).defaultValue("").required(false).build());
-    globalOperationParams.add(new ParameterBuilder().name("keys")
-        .description(ParameterDescriptions.DEPRECATED_USE_FILTER).modelRef(new ModelRef(string))
-        .parameterType(query).defaultValue("").required(false).build());
-    globalOperationParams.add(new ParameterBuilder().name("values")
-        .description(ParameterDescriptions.DEPRECATED_USE_FILTER).modelRef(new ModelRef(string))
-        .parameterType(query).defaultValue("").required(false).build());
```
joker234, line 185: This would remove these parameters from the Swagger visualization, but not from the swagger JSON? I'd prefer to keep them (as deprecated) for consistency. Otherwise this would be a breaking change, imho.

joker234: or: if you want to change them to “hidden” parameters, I would remove them from the docs as well

tyrasd: we'll hide them from the docs in a future release

joker234 (approved):

PR #171, opened 2021-03-31
add info regarding "NaN" ratio to docs

### LINKED ISSUE #166: NaN result is a string in JSON result
When you get `NaN` as result of an aggregation (e.g. through division by zero), the result is formatted as string (`"NaN"`). This is, because the [JSON standard](https://tools.ietf.org/html/rfc8259#section-6) doesn't allow you to return something like `NaN` as number.

I see two possibilities here:
* declare this as expected behaviour and write this behaviour in the docs
* return `null` in the JSON response

Example Request:
```bash
% curl -X POST https://api.ohsome.org/v1/elements/count/ratio \
    --data-urlencode "bboxes=0,0,0.01,0.01" \
    --data-urlencode "filter=geometry:polygon and amenity=hospital" \
    --data-urlencode "filter2=geometry:polygon and beds=* and amenity=hospital"
    --data-urlencode "time=2020-08-01" \
    --data-urlencode "format=json"
{
  "attribution" : {
    "url" : "https://ohsome.org/copyrights",
    "text" : "© OpenStreetMap contributors"
  },
  "apiVersion" : "1.4.1",
  "ratioResult" : [ {
    "timestamp" : "2020-08-01T00:00:00Z",
    "value" : 0.0,
    "value2" : 0.0,
    "ratio" : "NaN"
  } ]
}
```

The CSV format isn't affected as most CSV descriptions (e.g. [RFC 4180](https://tools.ietf.org/html/rfc4180)) don't have data types.

FabiKo117:
### Description
Ratio results can contain the value `"NaN"` as a string representation. This PR adds an info regarding that to the docs.

### Corresponding issue
Closes #166 

### Checklist
~My code follows the [code-style](https://github.com/GIScience/ohsome-api/blob/master/CONTRIBUTING.md#code-style) rules, and I have checked on the [static analyses](https://jenkins.ohsome.org/job/ohsome-api/view/change-requests/)~
~I have commented my code~
~I have added sufficient unit and API tests~

### DIFF
--- a/docs/endpoints.rst
+++ b/docs/endpoints.rst
@@ -237,6 +237,8 @@ What is the density of restaurants with wheelchair access in Heidelberg?
    :query types2: Deprecated! use **filter2**
    :query values2: Deprecated! see **filter2**
    
+.. note:: The result of a **ratio request** may contain the value **"NaN"**, for example when dividing `0` by `0`. 
+   
 **Example request**:
 
 How many oneway streets exist within living_street streets in Heidelberg over time? And how many of them are oneway streets?
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@ Changelog
 
 ## 1.5.0-SNAPSHOT (current master)
 
+### Other Changes
+
+* add information regarding a potential "NaN" value for `ratio` results to docs ([#166])
+
+[#166]: https://github.com/GIScience/ohsome-api/issues/154
+
 
 ## 1.4.1
 


bonaparten (approved):

docs/endpoints.rst
```diff
@@ -237,6 +237,8 @@ What is the density of restaurants with wheelchair access in Heidelberg?
    :query types2: Deprecated! use **filter2**
    :query values2: Deprecated! see **filter2**
    
+.. note:: The result of a **ratio request** may contain the value **"NaN"**, for example when dividing `0` by `0`.
```
tyrasd, line 240: IMHO the wording here is not great: a) The use of "_for example_" suggests that there are other situations where this could also occur, but AFAIK there aren't. b) The verb "_dividing_" suggests that this is something the user has done or can control; it could be made clearer that this is a result of a calculation where such a division occurred.

> .. note:: The result of a **ratio request** may contain the value **"NaN"**, when the ratio calculation involves a division of zero by zero.

FabiKo117: Good point, that's not in general the only operation where `NaN` could be the result, but as we do a division here for the ratio, it is in fact in our case. It's a bit unlucky that your suggestion did not come earlier :D but feel free to make a quick PR for that change if you'd like to.

tyrasd: :arrow_right:  https://github.com/GIScience/ohsome-api/pull/172

Issue #158, opened 2021-03-22
update outdated endpoint visualization graph in docs


FabiKo117:
We have a graph within our docs that shows the different endpoints in a [graphical visualization](https://docs.ohsome.org/ohsome-api/v1/endpoint-visualisation.html). 

By now, it is a bit outdated as the /contributions endpoints that came with the last two minor releases are not included yet.

Issue #150, opened 2021-03-12
[documentation] enhance error / response code documentation


joker234:
Currently, we don't have any documentation what errors a user can expect. [http-response-status.rst](https://github.com/GIScience/ohsome-api/blob/bbfbdf33066065103de7da8f7fffa0d14fc92065/docs/http-response-status.rst) provides a good, brief documentation on the status codes. This can be enhanced with all(?) specific errors, the ohsome API is able to return (with the exception of 500 internal errors, that are caused by other components).

Examples:
 * The 404 should mention the case when the time/area is not covered.
 * 401 isn't supported at the moment, correct?

FabiKo117:
With the PR tackling this issue, we should also add a note to the 200 response code regarding the special case when an error occurs during the processing of a data-extraction request. So that the result is then a broken GeoJSON containing the error JSON object at the end.

PR #101, opened 2021-01-21
add parameter description in aggregation responses

bonaparten:
### Description

Add descriptions of the response parameters in the docs.

### Corresponding issue

#88 

### New or changed dependencies

 none

### Checklist
- [ ] My code follows the [code-style](https://github.com/GIScience/ohsome-api/blob/master/CONTRIBUTING.md#code-style) rules, and I have checked on the [static analyses](https://jenkins.ohsome.org/job/ohsome-api/view/change-requests/)
- [ ] I have commented my code
- [ ] I have written javadoc (required for public methods)
- [ ] I have added sufficient unit and API tests
- [ ] I have made corresponding changes to the [documentation](https://github.com/GIScience/ohsome-api/tree/master/docs)
- [ ] I have updated the [CHANGELOG.md](https://github.com/GIScience/oshdb/blob/master/CHANGELOG.md)
- [ ] I have adjusted the examples in the [check-ohsome-api](https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/helpers/check-ohsome-api) script or [created an issue](https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/helpers/check-ohsome-api/-/issues/new) in the corresponding repository. More Information [here](https://github.com/GIScience/ohsome-api/blob/master/CONTRIBUTING.md#check-examples).

PR #194, opened 2021-05-12
Update documentation

### LINKED ISSUE #88: Extend docs on response parameters
- [ ] docs on different aggregation responses

bonaparten:
### Description

- Add descriptions of the response parameters for aggregation requests
- Add a list of the possible messages related to the HTTP response codes that the ohsome API can return
- Update the endpoints graph

### Corresponding issue

Closes #88, #150 and #158 

### Checklist
- [x] #101 needs to be closed


docs/endpoints.rst
```diff
@@ -840,28 +811,38 @@ Compare length of different types of streets for two or more regions.
...
+   :query types2: Deprecated! use **filter2**
+   :query values2: Deprecated! see **filter2**
 
 Users Aggregation
 -----------------
 
 .. http:post :: /users/count
 
-    Compute data aggregation functions on users. Possbile endpoints:
+    Get ``aggregation`` on OSM users. Possible endpoints:
```
tyrasd, line 831: To _get aggregation on users_ doesn't sound too clear to me. Maybe this would be better:

```suggestion
    Get ``aggregation`` statistics about OSM users. Possible endpoints:
```

docs/http-response-status.rst
```diff
@@ -1,25 +1,86 @@
...
+* "The given timeout does not fit to its format. Please give one value in seconds and use a point as the decimal delimiter, if needed."
+* "The given timeout is too long. It has to be shorter than [timeout] seconds"
+* "You need to give at least two timestamps or a time interval for this resource."
+* "The given parameter [parameter] can only contain the values 'true', 'yes', 'false', or 'no'."
+* "Your provided boundary parameter (bboxes, bcircles, or bpolys) does not fit its format, or you defined more than one boundary parameter."
+* "The geometry of your given boundary input could not be parsed for the creation of the response GeoJSON."
+* "The keys, values and types parameters must be empty, when you set the filter parameter."
+* "The properties parameter of this resource can only contain the values 'tags' and/or 'metadata' and/or 'contributionTypes' and/or 'unclipped'."
+* "The properties parameter of this resource can only contain the values 'tags' and/or 'metadata' and/or 'unclipped'."
+* "Parameter 'types' (and 'types2') can only have 'node' and/or 'way' and/or 'relation' OR 'point' and/or 'line' and/or 'polygon' and/or 'other'"
```
tyrasd, line 57: I don't know how useful a list of possible messages is, to be honest. Since they are just listed here plainly without further explanation, there's not really much users could learn from this list, could they? If anything, the list would be quite cumbersome to maintain (i.e. keep up-to-date) for us.

Instead one could simply write that: _“The error message contains further details about why the query failed.”

Or what was your reasoning behind including it? Maybe I'm simply not seeing it at the moment.

tyrasd (commented):
See some inline comments.

PS: can we shrink the canvas of the new [api tree graphic](https://github.com/GIScience/ohsome-api/blob/d32a9d699f7f78900b8920cc13a856fb3ba320f7/docs/_static/ohsome-api-tree.svg) to remove the extra white space at the top?

docs/http-response-status.rst
bonaparten: I understood #150 in that way. @joker234 Am I wrong? If yes, no problem, I can change it summarizing the reasons for every  HTTP response code of the ohsome API.

tyrasd: hm, ok. I would say it might be better to mention the possible cases in a slightly abstracted way, instead of a plain verbatim list. e.g. something like this:

```
* ``400 Bad Request`` - the ohsome API cannot or will not process the request due to a client error. This can occur because a parameter is not given in the correct format,  mandatory parameters are not present or other malformed requests. The error message contains further details about why the query failed.
* ``404 Not Found``-  the requested resource could not be found. This can occur when an invalid resource is requested, or when the provided boundary or time parameter does is not covered by the underlying data OSHDB database.
* ``413 Payload Too Large`` - the request is larger than the ohsome API is willing or able to process. Occurs when the given query took takes too long to compute.
```

joker234: I'm not quite sure what the intention should be. I would agree it doesn't make sense to joust copy & paste all error messages as you have to keep track of them, when you change the ohsome API's code. But maybe a little bit better explanation in the ohsome API documentation would be nice. Alternatively one could think about having more actual error codes in one list as the ORS has. https://github.com/GIScience/openrouteservice/blob/master/openrouteservice/docs/services/error_codes.md

bonaparten: I have updated this part trying to improve the explanation without the actual messages.

docs/http-response-status.rst
```diff
@@ -1,25 +1,27 @@
...
-* ``400 Bad Request`` - the ohsome API cannot or will not process the request due to a client error (e.g. malformed request syntax)
-* ``401 Unauthorized`` - the client does not have valid authentication credentials for the target resource
-* ``404 Not Found``-  the requested resource could not be found
-* ``405 Method not allowed`` - the requested method is not supported (e.g. a POST request for resources which accept only GET requests)
-* ``413 Payload Too Large`` - the request is larger than the ohsome API is willing or able to process
+``400 Bad Request`` - the ohsome API cannot process the request due to a client error. This can occur because a parameter is not given in the correct format, mandatory parameters are not present, unknown or not valid parameters are present, parameters are not unique, malformed GeoJSON, unsopported content-type reader, wrong parameter values or other malformed requests.
+
+``404 Not Found``-  the requested resource could not be found because the given boundary or time are not within the underlying osh-data.
+
+``413 Payload Too Large`` - the request is too large in respect to the given timeout parameter.
```
tyrasd, line 20: imho this could be worded a bit simpler/better? Specifically _"too large"_ means in our case that it _"took too long"_, and not all queries have a _"given timeout"_, but use the default one:

```suggestion
``413 Payload Too Large`` - the request took too long to compute.
```

bonaparten: OK, but it seems to me it doesn't inform the user in respect of what it took too long. What do you think about `the request took too long to compute in respect of the given or default timeout.`?

tyrasd: that should work

tyrasd (approved):

PR #198, opened 2021-05-26
Contributions latest count endpoints

### LINKED ISSUE #174: implement /contributions/latest/count
### Use Case Description
OQT would use this endpoint for the computation of the `lastEdit` indicator.

### Request Description
Enlarges the `/contributions` data-aggregation endpoint as an addition to the already existing `/contributions/latest/{geometry-type}` data-extraction endpoint.

bonaparten:
### Description
It adds the endpoint `/contributions/latest/count`, `/contributions/latest/count/density` and the new parameter `contributionType` for all the contributions aggregation endpoints. Through the new parameter, it is possible to filter contributions by `creation`, `deletion`, `geometryChange`, `tagChange` or a combination of them.

### Corresponding issue
Closes #174 

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ Changelog
 ### New Features
 
 * add new `contributionTypes` property to `properties` parameter ([#136])
+* add new `/contributions/latest/count` endpoint ([#174])
+* add new `contributionType` parameter for contributions aggregation ([#174])
 * contribution extraction endpoints now return the `@contributionChangesetId` as a separate field ([#200])
 
 ### Bug Fixes
@@ -15,8 +17,6 @@ Changelog
 * remove the parameters `@snapshotTimestamp` and `@lastEdit` from full-history extraction responses ([#191])
 
 ### Performance and Code Quality
-
-
 
 ### Other Changes
 
@@ -32,17 +32,16 @@ Changelog
 [#158]: https://github.com/GIScience/ohsome-api/issues/158
 [#164]: https://github.com/GIScience/ohsome-api/pull/164
 [#166]: https://github.com/GIScience/ohsome-api/issues/166
+[#174]: https://github.com/GIScience/ohsome-api/issues/174
 [#175]: https://github.com/GIScience/ohsome-api/issues/175
 [#184]: https://github.com/GIScience/ohsome-api/pull/184
 [#191]: https://github.com/GIScience/ohsome-api/issues/191
-
 
 ## 1.4.2
 
 * contribution extraction endpoints now return the `@contributionChangesetId` as a separate field (backported from 1.5.0) ([#200])
 
 [#200]: https://github.com/GIScience/ohsome-api/issues/200
-
 
 ## 1.4.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ Changelog
 
 [#200]: https://github.com/GIScience/ohsome-api/issues/200
 
+
 ## 1.4.1
 
 * upgrade OSHDB to version [OSHDB#0.6.4], fixes the following two bugs:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,10 +8,7 @@ Changelog
 * add new `contributionTypes` property to `properties` parameter ([#136])
 * add new `/contributions/latest/count` endpoint ([#174])
 * add new `contributionType` parameter for contributions aggregation ([#174])
-<<<<<<< HEAD
 * contribution extraction endpoints now return the `@contributionChangesetId` as a separate field ([#200])
-=======
->>>>>>> 5183d086 (chore: correct changelog)
 
 ### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ Changelog
 [#184]: https://github.com/GIScience/ohsome-api/pull/184
 [#191]: https://github.com/GIScience/ohsome-api/issues/191
 
+
 ## 1.4.2
 
 * contribution extraction endpoints now return the `@contributionChangesetId` as a separate field (backported from 1.5.0) ([#200])
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Changelog
 
 * add new `contributionTypes` property to `properties` parameter ([#136])
 * add new `/contributions/latest/count` endpoint ([#174])
+* add new `/contributions/latest/count/density` endpoint
 * add new `contributionType` parameter for contributions aggregation ([#174])
 * contribution extraction endpoints now return the `@contributionChangesetId` as a separate field ([#200])
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,7 @@ Changelog
 ### New Features
 
 * add new `contributionTypes` property to `properties` parameter ([#136])
-* add new `/contributions/latest/count` endpoint ([#174])
-* add new `/contributions/latest/count/density` endpoint
+* add new `/contributions/latest/count` and `/contributions/latest/count/density` endpoints ([#174])
 * add new `contributionType` parameter for contributions aggregation ([#174])
 * contribution extraction endpoints now return the `@contributionChangesetId` as a separate field ([#200])
 
@@ -37,7 +36,6 @@ Changelog
 [#175]: https://github.com/GIScience/ohsome-api/issues/175
 [#184]: https://github.com/GIScience/ohsome-api/pull/184
 [#191]: https://github.com/GIScience/ohsome-api/issues/191
-
 
 ## 1.4.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,10 @@ Changelog
 * add new `contributionTypes` property to `properties` parameter ([#136])
 * add new `/contributions/latest/count` endpoint ([#174])
 * add new `contributionType` parameter for contributions aggregation ([#174])
+<<<<<<< HEAD
 * contribution extraction endpoints now return the `@contributionChangesetId` as a separate field ([#200])
+=======
+>>>>>>> 5183d086 (chore: correct changelog)
 
 ### Bug Fixes
 
--- a/src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/ContributionsExecutor.java
+++ b/src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/ContributionsExecutor.java
@@ -90,7 +90,7 @@ public class ContributionsExecutor extends RequestExecutor {
           inputProcessor.getRequestUrlIfGetRequest(servletRequest));
     }
     if ("csv".equalsIgnoreCase(requestParameters.getFormat())) {
-      ExecutionUtils exeUtils = new ExecutionUtils(processingData);
+      var exeUtils = new ExecutionUtils(processingData);
       exeUtils.writeCsvResponse(results, servletResponse,
           ExecutionUtils.createCsvTopComments(URL, TEXT, Application.API_VERSION, metadata));
       return null;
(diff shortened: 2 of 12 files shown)


bonaparten:
I chose the name *otherChanges* for contributions that are neither changes of the tag nor changes of the geometry. I am open to better wording if you have some suggestions.

The `ContributionsExecutor` takes now a boolean more. I don't like to check boolean variables in this way. I think we should create inheritance between the executor classes (`/users/count` calls the contribution executor) and check by instance, but if I do that it would be nice to refactor all of the executor classes (so not only the contributions endpoints). I would rather do that all together with a bigger refactoring in the future. If you don't agree, I am open to refactor che contributions executor classes now.

src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/ContributionsExecutor.java
```diff
@@ -66,13 +100,87 @@ public Response count(boolean isUsersRequest) throws Exception {
...
+    if (contributionType == null) {
+      return mapRed;
+    }
+    List<String> contrTypes =
+        Arrays.asList("CREATION", "DELETION", "GEOMETRYCHANGE", "TAGCHANGE", "OTHERCHANGES");
+    contributionType = contributionType.toUpperCase();
+    if (!contrTypes.contains(contributionType)) {
+      throw new BadRequestException(
+          "The contribution type must be 'creation', 'deletion', 'geometryChange', 'tagChange' "
+              + "or 'otherChanges'.");
```
tyrasd, line 174: I think it would be a good addition to allow filtering for not only by a single contribution type, but by a list/set of types, e.g. it could look like this: `/contributions/count?…&contributionType=creation,deletion`. What do you think?

src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/ContributionsExecutor.java
```diff
@@ -66,13 +100,87 @@ public Response count(boolean isUsersRequest) throws Exception {
...
+    }
+    if (contributionType.equals("OTHERCHANGES")) {
+      return mapRed.filter(contr -> contr.getContributionTypes().isEmpty());
+    } else {
+      var string = new StringBuilder(contributionType);
+      if (contributionType.equals("TAGCHANGE") || contributionType.equals("GEOMETRYCHANGE")) {
+        string.insert(contributionType.length() - 6, '_');
+      }
+      return mapRed.filter(contr -> contr.is(ContributionType.valueOf(string.toString())));
```
tyrasd, line 183: not sure I like this "creative" implementation too much. I would have probably done it like this:

```
      contributionType = contributionType.toUpperCase();
      if (contributionType.equals("OTHERCHANGES")) {
        return mapRed.filter(contr -> contr.getContributionTypes().isEmpty());
      }
      ContributionType filterType;
      switch (contributionType) {
        case "CREATION":
          filterType = ContributionType.Creation;
          break;
        …
        default:
          throw new BadRequestException("…");
      }
      return mapRed.filter(contr -> contr.is(filterType));
```

IMHO it is more easily readable/understandable. Also it has the benefit that it's not dependent that the input strings correspond to the `ContributionType` enum and avoids potentially thousands of calls to `ContributionType.valueOf()`.

(just a suggestion)

tyrasd (changes requested):
Please find some comments/remarks inline.

Generally, I'm wondering if it's a good idea to include the `other` case for the `contributionType` parameter/"filter". The reason is that a) I can't think of a good use case for this, b) the implementation is very slow performance wise, because it needs to always check for all contribution types (especially the _geometry change_ check can be very expensive).

src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/ContributionsExecutor.java
```diff
@@ -32,22 +44,44 @@ public ContributionsExecutor(HttpServletRequest servletRequest,
...
+   *         #contributionsCount(MapReducer, boolean) contributionsCount}
    */
-  public Response count(boolean isUsersRequest) throws Exception {
-    MapReducer<OSMContribution> mapRed = inputProcessor.processParameters();
+  public Response count(boolean isUsersRequest, boolean isContributionsLatestCount)
+      throws UnsupportedOperationException, Exception {
+    MapReducer<OSMContribution> mapRed;
     final SortedMap<OSHDBTimestamp, ? extends Number> result;
+    if (isContributionsLatestCount) {
+      inputProcessor.getProcessingData().setFullHistory(true);
```
tyrasd, line 71: why is this necessary (only for the _latest_ resource)? IMHO it's not obvious, so it should be documented here with an inline comment.

src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/ContributionsExecutor.java
```diff
@@ -66,13 +100,87 @@ public Response count(boolean isUsersRequest) throws Exception {
...
+      boolean isContributionsLatest) throws UnsupportedOperationException, Exception {
+    if (isContributionsLatest) {
+      MapReducer<List<OSMContribution>> mapRedGroupByEntity = mapRed.groupByEntity();
+      Optional<FilterExpression> filter = processingData.getFilterExpression();
+      if (filter.isPresent()) {
+        mapRedGroupByEntity = mapRedGroupByEntity.filter(filter.get());
+      }
+      return contributionsFilter(mapRedGroupByEntity
+          .map(listContrsPerEntity -> listContrsPerEntity.get(listContrsPerEntity.size() - 1)))
+          .aggregateByTimestamp(OSMContribution::getTimestamp).count();
```
tyrasd, line 151: Can you explain why the `aggregateByTimestamp` is done after the `latest` contribution is determined from the list of contributions?

I mean, the exact behaviour of the endpoint is not specified in the documentation[¹], but intuitively I would have expected that if one requests multiple time intervals, that each one is handled individually. i.e. that the output of `/contributions/count/latest?…&time=t1,t2,t3` is the "concatenation" of the out two requests with `time=t1,t2` and `time=t2,t3`. The current implementation would be instead "distribute" the total result (i.e. `time=t1,t3`) into the two intervals from `time=t1,t2,t3`.

[¹] maybe we should do that?

src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/ContributionsExecutor.java
bonaparten: Regarding this point:
1. I think the Inputprocessor is a little bit outdated. With the whole bunch of endpoints that we have by now, the check of `isFullHistory()` (and some other checks) is not intuitive. Some future refactoring would improve the readability.
2. The documentation for groupByEntity() in the oshdb API refers only to map() and flatMap(). It was not intuitive that also calling filter() prevents the call of groupByEntity(). Is there some room for docs improvement there?

src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/ContributionsExecutor.java
bonaparten: I considered the endpoint with respect to entities (latest among all contributions of an entity in the given interval). Otherwise, given one entity having some contributions in two time intervals, the output will have the latest contribution in the first interval for the entity and another latest contribution for the same entity in the second interval.

CHANGELOG.md
```diff
@@ -32,18 +32,17 @@ Changelog
 [#158]: https://github.com/GIScience/ohsome-api/issues/158
 [#164]: https://github.com/GIScience/ohsome-api/pull/164
 [#166]: https://github.com/GIScience/ohsome-api/issues/166
+[#174]: https://github.com/GIScience/ohsome-api/issues/174
 [#175]: https://github.com/GIScience/ohsome-api/issues/175
 [#184]: https://github.com/GIScience/ohsome-api/pull/184
 [#191]: https://github.com/GIScience/ohsome-api/issues/191
 
-
```
tyrasd, line 39: In the past we always put in an extra newline before the `##` heading of a new version. I'd say it would be best to use a consistent style throughout the document.

src/test/java/org/heigit/ohsome/ohsomeapi/controller/GetControllerTest.java
```diff
@@ -681,6 +681,76 @@ public void getUsersCountDensityGroupByBoundaryTest() {
...
+        + "/contributions/latest/count?bboxes=8.673088,49.401834,8.692051,49.407979&"
+        + "filter=type:way and building=residential&format=json&time=2016-01-01/2019-01-01/P1Y&"
+        + "contributionType=tagChange", JsonNode.class);
+    int sum = StreamSupport
+        .stream(Spliterators.spliteratorUnknownSize(response.getBody().get("result").iterator(),
+            Spliterator.ORDERED), false)
+        .mapToInt(node -> node.get("value").asInt()).sum();
+    assertEquals(4, sum);
+  }
+
```
tyrasd, line 739: checkstyle complains about this line:

```suggestion

```

src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/ContributionsExecutor.java
```diff
@@ -66,13 +95,98 @@ public Response count(boolean isUsersRequest) throws Exception {
...
+      }
+    }
+    List<ContributionType> contributionTypes = new ArrayList<>();
+    givenTypes.stream().forEach(givenType -> {
+      for (ContributionType contrType : ContributionType.values()) {
+        if (contrType.name().equals(givenType)) {
+          contributionTypes.add(contrType);
+        }
+      }
+    });
```
tyrasd, line 197: I strongly advice to use static typing wherever possible. Here you are circumventing the compiler's ability to check types (specifically the `ContributionType` enum) by looping through the enum values and comparing them to strings. IMHO this does not provide any benefit over something simple like this:

```java
    List<String> contributionTypes = new ArrayList<>();
    for (String givenType : types.split(",")) {
      switch (givenType) {
        case "CREATION":
          contributionTypes.add(ContributionType.CREATION);
          break;
        case "DELETION":
          contributionTypes.add(ContributionType.DELETION);
          break;
        case "TAGCHANGE":
          contributionTypes.add(ContributionType.TAG_CHANGE);
          break;
        case "GEOMETRYCHANGE":
          contributionTypes.add(ContributionType.GEOMETRY_CHANGE);
          break;
        default:
          throw new BadRequestException("The contribution type must be 'creation', 'deletion',"
              + "'geometryChange', 'tagChange' or a combination of them");
      }
    }
```

src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/ContributionsExecutor.java
```diff
@@ -1,12 +1,21 @@
+
```
tyrasd, line 1: please remove this (accidentally committed?) newline here

tyrasd (changes requested):
See the inline comments and the following general comment below. Hopefully these will be the last ones.

I'm wondering if it would make sense to also implement a `/contributions/latest/count/density` endpoint. I guess we don't immediately have a feature request for it, but IMHO it would make sense to implement it as well, since all(?) other `count` requests also provide the `density` variant :thinking:

src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/ContributionsExecutor.java
tyrasd: ad 2.: yes, you are right. some filters also act like a `flatMap` in the OSHDB. This is indeed missing/outdated information in the OSHDB docs. Can you [open a ticket](https://github.com/giscience/oshdb/issues/new) for that in the OSHDB repo? Thanks!

src/test/java/org/heigit/ohsome/ohsomeapi/controller/GetControllerTest.java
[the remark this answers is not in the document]
tyrasd: (sorry, i saw that this was already fixed in the meantime, but I couldn't remove it from the review comments)

tyrasd (approved):

PR #202, opened 2021-06-10
Check-ohsome-api should be run at every PR

bonaparten:
It makes clear that [check-ohsome-api](https://gitlab.gistools.geog.uni-heidelberg.de/giscience/big-data/ohsome/helpers/check-ohsome-api) should be run at every pull request.


joker234 (dismissed):

bonaparten:
@tyrasd sure :-D. Do you think it is not necessary?

PR #222, opened 2021-07-26
upgade to OSHDB version 0.7.0

tyrasd:
### Description
upgrades to OSHDB version 0.7

still to do:
* [x] look for usages of (now) deprecated methods and replace
* [x] upgrade test file on Jenkins (CI) to 0.7 db schema
* [x] check for stuff to update in readme / docs(?)
* [x] add test file to downloads server and update link in readme

### New or changed dependencies
- upgrade oshdb to 0.7.0
- add oshdb-api-ignite

### Checklist


tyrasd:
> look for usages of (now) deprecated methods and replace

AFAICT, the only remaining places of uses of `osmType` and `osmTag` are in `InputProcessor.java` in code paths which are already deprecated there already (i.e. belonging to the old `types/keys/values` parameters). I'd not hassle to upgrade those to the new filters.

tyrasd:
>  check for stuff to update in readme / docs(?)

I haven't found anything relevant in the docs, but the download link for the test file needs to be adjusted in the project readme

README.md
```diff
@@ -129,7 +129,7 @@ To run the tests locally, you need the following:
     * -Dport.get starts data-aggregation + metadata tests using GET requests
     * -Dport.post starts data-aggregation tests using POST requests
     * -Dport.data starts data-extraction tests using GET and POST requests
-2. [heidelberg.oshdb](https://downloads.ohsome.org/OSHDB/v0.6/europe/germany/baden-wuerttemberg/heidelberg_68900_2020-07-23.oshdb.mv.db) file (or any other, which includes the data from Heidelberg)
+2. [heidelberg.oshdb](https://downloads.ohsome.org/OSHDB/v0.7/europe/germany/baden-wuerttemberg/heidelberg.oshdb.mv.db) file (or any other, which includes the data from Heidelberg)
```
rtroilo, line 132: This download file does not exist yet. I will prepare an up-to-date extract for heidelberg.
Should we include in to the filename the version like
- heidelberg_v0.7.oshdb.mv.db

pom.xml
```diff
@@ -93,6 +93,21 @@
...
+        <exclusion>
+          <groupId>org.slf4j</groupId>
+          <artifactId>slf4j-simple</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.slf4j</groupId>
+          <artifactId>slf4j-log4j12</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
```
rtroilo, line 110: This is write, but what do you think to remove the dependency of 
- sl4j-simple
from the oshdb-api. So that we don't need to exclude this in such an situation.
We should create a new issue for that.

src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/ExecutionUtils.java
```diff
@@ -600,10 +597,10 @@ public void writeCsvResponse(Result[] resultSet, HttpServletResponse servletResp
...
-    for (int i = 0; i < tags.length; i += 2) {
-      int tagKeyId = tags[i];
-      int tagValueId = tags[i + 1];
+    Iterable<OSHDBTag> tags = f.getEntity().getTags();
+    for (OSHDBTag tag : tags) {
+      int tagKeyId = tag.getKey();
+      int tagValueId = tag.getValue();
       if (tagKeyId == keysInt) {
         if (valuesInt.length == 0) {
           return new ImmutablePair<>(new ImmutablePair<>(tagKeyId, tagValueId), f);
```
rtroilo, line 606: Why not just take the `OSHDBTag` here instead of creating a new `ImmutablePair(tagKeyId,  tagValueId)`.

src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/UsersRequestExecutor.java
```diff
@@ -129,10 +130,10 @@ public static Response countGroupByTag(HttpServletRequest servletRequest,
...
-      for (int i = 0; i < tags.length; i += 2) {
-        int tagKeyId = tags[i];
-        int tagValueId = tags[i + 1];
+      Iterable<OSHDBTag> tags = ExecutionUtils.extractContributionTags(f);
+      for (OSHDBTag tag : tags) {
+        int tagKeyId = tag.getKey();
+        int tagValueId = tag.getValue();
         if (tagKeyId == keysInt) {
           if (valuesInt.length == 0) {
             res.add(new ImmutablePair<>(new ImmutablePair<>(tagKeyId, tagValueId), f));
```
rtroilo, line 139: Same here, we could use `OSHDBTag tag` here instead of creating a new Pair of `new ImmutablePair<>(tagKeyId, tagValueId)`.

rtroilo (changes requested):
Looks very good, thank you.
I just got some questions.

README.md
tyrasd: IMHO it's good enough to have the version number in the download path.

pom.xml
tyrasd: >  We should create a new issue for that.

:+1: yeah, this would be an issue for the oshdb repo.

src/main/lombok/org/heigit/ohsome/ohsomeapi/executor/ExecutionUtils.java
tyrasd: makes sense, will do :+1:

tyrasd: well, actually, the values returned by this method are later required to be Comparable, which `OSHDBTag` currently isn't. Since fixing that wouldn't be a trivial change, I'd rather leave the current implementation as it is, and deal with it later.

rtroilo (approved):

PR #233, opened 2021-09-09
Error message for requests without defined parameters

### LINKED ISSUE #227: specific error message for POST/GET requests with empty request body/search-parameters
When performing a POST request with an empty request body (i.e. without specifying any parameter), or analogously a GET request without any search parameters, the ohsome api produces currently an error message which is a bit too specific and does not help perfectly to identify the root cause of the issue:

> You need to define one of the boundary parameters (bboxes, bcircles, bpolys).

It would be better to error with a message like the following (to be tweaked!):

> The query did not specify any parameters (e.g. an empty request body), please check your code.

bonaparten:
It gives back an error message in the case of queries without defined parameters.

### Corresponding issue
Closes #227 

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ Changelog
 ### New Features
 
 * add `contributions/count[/density]/groupBy/boundary` endpoints to fetch contribution counts/densities grouped by boundary ([#217])
-* add an error message for requests without defined parameters. [#233]
+* add an error message for requests without defined parameters ([#233])
 
 ### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Changelog
 ### New Features
 
 * add `contributions/count[/density]/groupBy/boundary` endpoints to fetch contribution counts/densities grouped by boundary ([#217])
+* add error message in cases of queries without defined parameter [#233]
 
 ### Bug Fixes
 
@@ -15,7 +16,7 @@ Changelog
 [#217]: https://github.com/GIScience/ohsome-api/issues/217
 [#228]: https://github.com/GIScience/ohsome-api/pull/228
 [#230]: https://github.com/GIScience/ohsome-api/pull/230
-
+[#233]: https://github.com/GIScience/ohsome-api/issues/227
 
 ## 1.6.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ Changelog
 ### New Features
 
 * add `contributions/count[/density]/groupBy/boundary` endpoints to fetch contribution counts/densities grouped by boundary ([#217])
-* add an error message for requests without defined parameters ([#233])
+* add an error message for requests without defined parameters ([#227])
 
 ### Bug Fixes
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -15,7 +16,7 @@ Changelog
 [#217]: https://github.com/GIScience/ohsome-api/issues/217
 [#228]: https://github.com/GIScience/ohsome-api/pull/228
 [#230]: https://github.com/GIScience/ohsome-api/pull/230
-
+[#233]: https://github.com/GIScience/ohsome-api/issues/227
```
tyrasd, line 19: let' use the ticket number for the markdown link: `[#227]` instead of ~`[#233]`~. This needs to be changed above in line 9 as well.

src/test/java/org/heigit/ohsome/ohsomeapi/controller/PostControllerTest.java
```diff
@@ -112,9 +112,20 @@ public void nonNodedLinestringsIntersectionTest() {
...
   /*
-   * false parameter tests
+   * false parameter and no parameters tests
    */
 
+  @Test
+  public void queryWithoutParametersTest() {
+    TestRestTemplate restTemplate = new TestRestTemplate();
+    ResponseEntity<JsonNode> response =
+            restTemplate.postForEntity(server + port + "/elements/perimeter", null,
```
tyrasd, line 122: the indentation here is different from e.g. [line 135 below](https://github.com/GIScience/ohsome-api/pull/233/files#diff-5738da37ce58a90b3ab106a9ff11247cf4cad7fa136f4fc45233a94d5dedbc08R135). Maybe the rest of the statement (`JsonNode.class);`) would otherwise fit into a single line?

src/main/lombok/org/heigit/ohsome/ohsomeapi/inputprocessing/InputProcessor.java
```diff
@@ -709,13 +709,18 @@ private void checkContentTypeHeader(HttpServletRequest servletRequest) {
...
+   * fuzzy matching scores.
    *
-   * @throws BadRequestException in case of invalid parameter or if a parameter is given more than
-   *         once
+   * @throws BadRequestException in case of no parameters, invalid parameters or if parameters are
+   *     given more than once.
    */
   private void checkParameters(HttpServletRequest servletRequest) {
+    if (servletRequest.getParameterMap().isEmpty()) {
+      throw new BadRequestException("The query did not specify any parameter. Please remember: "
```
tyrasd, line 721: maybe we can store this message also into the `ExceptionMessages` class?!

tyrasd (commented):
see inline comments

CHANGELOG.md
bonaparten: My bad, I missaw the issue number.

src/main/lombok/org/heigit/ohsome/ohsomeapi/inputprocessing/InputProcessor.java
bonaparten: The message is present in this line and in the corresponding test method, I don't think we will use this message in the future somewhere else, but yes I can add it.

tyrasd: hm, the benefit would be that we would have one central place where to find and adjust error messages if needed.

we have quite a few already in there which are only used once (excluding tests), e.g. `BOUNDARY_PARAM_GEOJSON_FORMAT`, `BPOLYS_PARAM_GEOMETRY`, `BOUNDARY_IDS_FORMAT`, …, `TIME_FORMAT_CONTRIBUTION`, `TIMEOUT`, `TIMEOUT_FORMAT` – what would you do about these then?

bonaparten: Yes, but we are a little bit choppy regarding this point since there are also error messages not stored in variables. But if this is in general the right way to do it, I will do it.

tyrasd:
store full error message in constant

tyrasd (approved):
FYI: I made some additonal tweaks in 93744b5d905f1e584104811cb177a1d4d9b01761, looks good to me now. :+1:

tyrasd (approved):

PR #291, opened 2023-01-11
fix bug in …/groupBy/tag and …/groupBy/key endpoints when key/values do not exist in keytables

tyrasd:
Fixes a bug/crash in `…/groupBy/tag` requests triggered when using non-existing `groupByKey` or `groupByValues` are used in a query.

### To Reproduce

```
curl -X GET "https://api.ohsome.org/v1/elements/count/groupBy/tag?bboxes=8.67%2C49.39%2C8.71%2C49.42&filter=type%3Away%20and%20building%3D*&format=json&groupByKey=building&groupByValues=DoesNotExist1&time=2023-01-01" -H  "accept: application/json"
```

This query is requesting results grouped by the tag `building` and value `DoesNotExist1` (which does not exits in the keytables). The result, however contains `building=yes` as the `groupByObject`. When performing a query with two non-existing `groupByValues` (e.g. `DoesNotExist1,DoesNotExist2`), the results includes two entries with `remainder` as `groupByObject`. When requesting three or more non-existing `groupByValues`, the query crashes and a 500 server error is returned.

Similar bugs occur when performing a query with a non-existing tag key as the `groupByKey`.

### The Problem and Fix

The issue is caused by a mistake in the code, where consecutive negative tag numbers are used as `keyInt`/`valuesInt` for OSM tags which do not exist in the keytables, but the hardcoded value `-1` is also used to represent the `remainder` portion of the result. This is causing the mixup between the value `DoesNotExist1` and the superfluous `remainder` entry in the example result above. Further, the negative pseudo-indices started at 0 which is also the index of the most common tag value for the corresponding tag. This is causing the mixup between the value `DoesNotExist2` and the result entry for the value `yes` in the example above. The crashes for any additional non-existing values are caused by a null pointer exception where the (still) non-existing pseudo-indices are attempted to be converted back to strings using the tag translator.

This is fixed by starting the negative pseudo-indices at position -2 where confusion with the -1 value for the `remainder` entry is possible and using the input values to resolve the pseudo-indices back to strings.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@ Changelog
 
 ## 1.9.0-SNAPSHOT (current master)
 
+### Bug Fixes
+
+* Fix crash and incorrect output of `…/groupBy/tag` and `…/groupBy/key` endpoints when non-existing tag keys or values are used in a query ([#291])
+
+[#291]: https://github.com/GIScience/ohsome-api/pull/291
+
 
 ## 1.8.0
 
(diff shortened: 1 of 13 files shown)


rtroilo (approved):
Thank you

PR #297, opened 2023-03-08
Clarify what the groupBy* parameters actually do

### LINKED ISSUE #296: Improve wording of documentation for groupByKeys
### Use Case Description

`:query groupByKeys: OSM key(s) given as a list and combined with the ‘AND’ operator; default: empty;` in https://github.com/GIScience/ohsome-api/blob/master/docs/group-by.rst#L28 is difficult to understand (and possibly wrong?). I don't think there is a need to be this short.

### Request Description

Suggested update: 

`:query groupByKeys: OSM key(s) given as a comma separated list like "key1,key2...". default: empty;` 

The description should additionally contain the following text, but that may be put into the main description of the function:

`The result will contain separate results for each key and the 'remainder' meaning all objects that fulfilled the filter but had none of the 'groupByKeys'`.

### Additional Information

Similar text also occurs in other locations.

//cc @redfrexx

tyrasd:
Improve documentation by replacing old (copy-pasted?) parameter descriptions with how they actually are working.

Closes #296

### DIFF
--- a/docs/group-by.rst
+++ b/docs/group-by.rst
@@ -25,7 +25,9 @@ Groups the result by the given keys that are defined through the ``groupByKeys``
 
     :param aggregation: aggregation type, one of ``area``, ``count``, ``length``, ``perimeter``
     :query <params>: see query-parameters_ at /elements(aggregation) endpoint
-    :query groupByKeys: OSM key(s) given as a list and combined with the ‘AND’ operator; default: empty;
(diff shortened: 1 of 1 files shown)


joker234 (approved):

joker234 (approved):

rtroilo (approved):
thank you

PR #312, opened 2023-09-29
upgrade to OSHDB v1.2.1 and prepare v1.10.1 release

tyrasd:
for the fix in https://github.com/GIScience/oshdb/pull/516


rtroilo (approved):
thank you

